PR #25, opened 2016-07-19
Modularized project card & enabled details view

mmmavis:
- Modularized project card (relates to #16)
  - extracted `ProjectCard` from `pulse.js` and put it to a new `project-card.js` file. This helps improve code readability. If we are ever going to "reactify" network pulse this extraction work would the "reactification" process smoother.

- Enabled details view (relates to #14)
 - just wanted to get the feature in place first. we will keep iterating on code & design.

PR #36, opened 2016-07-20
Notifier module

### LINKED ISSUE #29: Move notifications to separate module
https://github.com/mozilla/network-pulse/blob/gh-pages/js/pulse.js#L278

mmmavis:
Fixes #29 

@acabunoc this is similar to what I did in PR https://github.com/mozilla/network-pulse/pull/25. 

- 1st commit https://github.com/mozilla/network-pulse/commit/40a18092d66ea6b92a9868d2b583580e9387eba3 is just changing the way the object is created (using object literal instead)
- 2nd commit https://github.com/mozilla/network-pulse/commit/a1b3e50eaa6e1f833f812355fbff72fb66bb2995 I renamed the object `notify` to `Notifier`. And changed occurrences of `Notifier` within `Notifier` itself to `this`.

### DIFF
--- a/js/notifier.js
+++ b/js/notifier.js
@@ -45,20 +45,17 @@ var Notifier = {
       }   
     }
   },
-  checkForUpdates: function(newestTimestamp,newestTitle) {
+  checkForUpdates: function(sortedProjects) {
     var updated = false;
+    var newestProject = sortedProjects[0];
+    var newestTimestamp = Date.parse(newestProject.Timestamp);
     var lastProject = Number(localStorage.getItem("lastProject")); 
     if (lastProject && lastProject < newestTimestamp) { 
-      var notification = 'Check out ' + newestTitle;
+      var notification = 'Check out ' + newestProject.Title;
       this.create(notification);
       updated = true;
+      localStorage.setItem("lastProject",newestTimestamp);
     }
-    localStorage.setItem("lastProject",newestTimestamp);
     return updated;
-  },
-  init: function(action,interval) {
-    if (action && typeof action === 'function') {
-      setInterval(action, interval || this.defaultInterval);
-    }
   }
 };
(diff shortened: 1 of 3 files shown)


mmmavis:
@acabunoc to check out the notification a quickest hack I found is just to comment out the `if (lastProject && lastProject < newestTimestamp) ` check in notifier.js (just the conditional check, not the entire block)

xmatthewx:
😸

acabunoc:
I'll take a look after the MLN call today!

js/notifier.js
```diff
@@ -0,0 +1,64 @@
...
+      var notification = 'Check out ' + newestTitle;
+      this.create(notification);
+      updated = true;
+    }
+    localStorage.setItem("lastProject",newestTimestamp);
+    return updated;
+  },
+  init: function(action,interval) {
+    if (action && typeof action === 'function') {
+      setInterval(action, interval || this.defaultInterval);
```
acabunoc: Just a thought:

It's not super clear that turning on notifications turns on auto-refresh -- having both in this file feels a bit messy, but makes sense if I know the two are related!

I would consider separating them a bit and moving this `setInterval` into `PulseMaker` since we're using it to call `PulseMaker.refresh` and leave this file to have just Notification logic. OR we could make it clearer in the Feature flags that notifications turn on the auto-refresh.

acabunoc:
Looks good! Left a thought, but it's minor :)

js/notifier.js
mmmavis: @acabunoc totally agree with you. There is still some tweaking work (especially on `PulseMaker`) that needs to be done... I just couldn't decide if I should file separate PRs or not. 😉 

I'll move `setInterval` to `PulseMaker` first and refining `PulseMaker` in follow-up PRs!

PR #118, opened 2016-08-05
"Project not found" screen

### LINKED ISSUE #73: Implement 404 broken link screen
@cadecairos brought this up while QAing a PR: https://github.com/mozilla/network-pulse/pull/50#issuecomment-236014187

> Yeah, we should consider this just a prototype - although I bet we'll get lots of feedback about it from our early adopters :).

> I also just realized I should try loading the page with a bogus project id in the query string. You end up with just the lightbox without content. Is there a better way of handling this?

As @kristinashu [suggested](https://github.com/mozilla/network-pulse/pull/50#issuecomment-236056940) for now we should

> just have a message and a link to the landing screen.

mmmavis:
Fixes #73

### DIFF
--- a/js/views-manager.js
+++ b/js/views-manager.js
@@ -11,8 +11,8 @@ var ViewsManager = {
   projects: [],
   $controlsContainer: $("#project-container .controls"),
   $projectsContainer: $("#project-container .projects"),
-  NotFoundView: {
-    _$container: $("#not-found-message"),
+  MessageView: {
+    _$container: $("#message-view"),
     setMessages: function(header,bodyText) {
       this._$container.find("h2").html(header);
       this._$container.find("p").html(bodyText);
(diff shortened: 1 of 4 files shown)


js/views-manager.js
```diff
@@ -11,6 +11,19 @@ var ViewsManager = {
   projects: [],
   $controlsContainer: $("#project-container .controls"),
   $projectsContainer: $("#project-container .projects"),
+  NotFoundView: {
```
acabunoc: Since you use this for:
* nothing found in search,
* broken direct link, and
* prompt for saving favourites

Maybe we should name this more generically than "NotFoundView"? Something like NotificationView or MessageView

acabunoc:
Minor comment on naming, otherwise this looks good! R+ 🍰

js/views-manager.js
acabunoc: Not NotificationView! Forgot we have something else called notifications...

mmmavis:
yay done! 👯 👯 👯

PR #124, opened 2016-08-08
Turned off auto refresh for now

mmmavis:
Fixes #122

PR #133, opened 2016-08-08
Fixed notification bug

### LINKED ISSUE #62: Fix update & notification logic
There's a bug in the auto updates. 

Steps to repeat: 
 - open pulse and dev console
 - add a new asset via form
 - wait for check. console reports `updated:  false` 
 - new project doesn't appear in the stream
 - notification isn't triggered

Note: we should fix this, but it is lower priority than other issues in this milestone.

mmmavis:
Fixes #62.

PR https://github.com/mozilla/network-pulse/pull/124 turns off auto refresh & updates (for now) while this PR fixes the currently broken notification logic

To test

- make sure `FEATURE.notify` in `pulse.js` is set to `true`
- you can lower `REFRESH_INTERVAL` if you are like me who's impatient 😉 
- try submitting a new project, return to the app, and see you get a desktop notification for that newly added project

### DIFF
--- a/js/notifier.js
+++ b/js/notifier.js
@@ -1,7 +1,7 @@
 /* notifications */
 
 var Notifier = {
-  defaultInterval: 1*60*1000,
+  LOCALSTORAGE_KEY: "pulse-newestProjectTimestamp",
   checkPermission: function() {
     var permission = false;
     if (Notification.permission === "granted") {
@@ -49,13 +49,18 @@ var Notifier = {
     var updated = false;
     var newestProject = sortedProjects[0];
     var newestTimestamp = Date.parse(newestProject.Timestamp);
-    var lastNewestProjectTimestamp = Number(localStorage.getItem("newestProjectTimestamp"));
-    if (lastNewestProjectTimestamp && lastNewestProjectTimestamp < newestTimestamp) {
+    var localStorageKey = this.LOCALSTORAGE_KEY;
+    var newestTimestampInStorage = localStorage.getItem(localStorageKey);
+    var onInitialLoad = !newestTimestampInStorage;
+    var hasUpdates = Number(newestTimestampInStorage) < newestTimestamp;
+
+    if ( onInitialLoad || hasUpdates ) {
       var notification = 'Check out ' + newestProject.Title;
       this.create(notification);
+      localStorage.setItem(localStorageKey,newestTimestamp);
       updated = true;
-      localStorage.setItem("newestProjectTimestamp",newestTimestamp);
     }
+
     return updated;
   }
 };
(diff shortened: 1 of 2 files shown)


js/notifier.js
```diff
@@ -49,13 +49,18 @@ var Notifier = {
     var updated = false;
     var newestProject = sortedProjects[0];
     var newestTimestamp = Date.parse(newestProject.Timestamp);
-    var lastNewestProjectTimestamp = Number(localStorage.getItem("newestProjectTimestamp"));
-    if (lastNewestProjectTimestamp && lastNewestProjectTimestamp < newestTimestamp) {
```
mmmavis: fixed a bug here - we forgot to check for the initial load and therefore `newestProjectTimestamp` never got saved to `localStorage` (i.e., this `if` condition we had before could never be met)

js/pulse.js
```diff
@@ -106,7 +106,6 @@ var newProjectForm = {
     newProjectForm.formContainer.style.display = 'none';
     newProjectForm.projectContainer.style.display = 'block';
     newProjectForm.toggleFormButton.style.transform = 'rotate(0deg)';
-    setTimeout(function(){ PulseMaker.getData(); }, 250);
```
mmmavis: no need to do this again since `refresh` is already a recurring task
(see `pulse.js`, the `PulseMaker.init` method)

mmmavis:
@xmatthewx @cadecairos @acabunoc review? 🍰 🍕 🍹

js/pulse.js
acabunoc: Can we update the comment here
>  this turns on both auto app refresh and browser notification

 https://github.com/mmmavis/network-pulse/blob/9c2ed73f25c92a0a2d10d281d617d35bbe07f465/js/pulse.js#L7

To only mention browser notifications?

acabunoc: Nevermind, just noticed the refresh is still wrapped in `if(FEATURE.notify)` :)

acabunoc:
So I can see the "updated: true" but no desktop notification. I granted permissions & ran this on a local server, so I'm not sure what I'm missing... @mmmavis ?

js/pulse.js
mmmavis: I will clean up the refresh & notification logic after we have decided how it should work 😉 We have a few tickets open for this discussion!

mmmavis:
Just tried it again on FF and it was working as expected to me. Not sure what's causing the problem... but since the notification feature is currently turned off, @acabunoc and I are good with merging this PR now and revisit this issue when we decide to turn the feature back on.

mmmavis:
🍸 🍸 🍸 🍸

PR #141, opened 2016-08-09
Added more metadata(`tags` `issues` `program` `type`)

### LINKED ISSUE #58: Add more metadata to Project
Based on https://github.com/mozilla/network-pulse/issues/33 and comps https://redpen.io/p/by8c797c413d39c620

When a user clicks on "Issues" the 5 Issue tags (Online Privacy & Security, Open Innovation, Decentralization, Web Literacy, Digital Inclusion) appear below with some intro text.

If a user clicks on one of the tags, then only posts with that specific tag are visible.

Only one tag can be selected at a time.

I've added a column to the spreadsheet (https://docs.google.com/spreadsheets/d/1vmYQjQ9f6CR8Hs5JH3GGJ6F9fqWfLSW0S4dz-t2KTF4/edit#gid=922457769) and assigned Issues to each project. Some don't have an issue assigned and some have multiple issues (separated by commas).

We will hold off on adding this field to the form.

@mmmavis let me know if you have any questions/comments

mmmavis:
Fixes #58 

Added more metadata(`tags` `issues` `program` `type`) to Project Card but do not make them visible on UI (see https://github.com/mozilla/network-pulse/issues/58#issuecomment-235935083)

Note that this PR is about making values of  `tags`, `issues`, `program`, and `type` searchable with the **current** search setup (i.e., search through `textContent` within Project Card)

/cc @acabunoc @cadecairos 

---

To test, you can temporary remove class `hidden-meta` from `index.html`. The additional metadata will then show on UI, within Project Card. Stay on the Features tab and search for "literacy", you should see 3 matching projects instead of 1 (<-- the result you will get if you try the same search term on https://mozilla.github.io/network-pulse/)

### DIFF
--- a/index.html
+++ b/index.html
@@ -91,7 +91,7 @@
         featured ? "featured" : "",
       ].join(" ");
     %>
-    <div id="<%= id %>" class="<%= classNames %>" data-id="<%= id %>" data-timestamp="<%= rawTimestamp %>">
+    <div id="<%- id %>" class="<%- classNames %>" data-id="<%- id %>">
       <% if (thumbnail) { %>
       <img src="<%- thumbnail %>" class="thumbnail" width="100%" />
       <% } %>
@@ -100,6 +100,7 @@
         <h3><%- creator %> on <%- timestamp %></h3>
         <p class="description"><%- description %></p>
         <p class="interest"><%- interest %></p>
+        <div class="hidden-meta"><%- tags %> <%- issues %> <%- program %> <%- type %></div>
         <div class="project-links">
           <div class="btn-group">
             <% if (link) { %>
--- a/js/project-card.js
+++ b/js/project-card.js
@@ -1,16 +1,9 @@
 /* Project Card */
 
 var ProjectCard = {
-  ID_PREFIX: "p",
   template: _.template($("script#project-card-template").html()),
-  getProjectCardId: function(projectData) {
-    var timestamp = projectData.Timestamp ? projectData.Timestamp : false;
-    var projectId = timestamp ? Date.parse(timestamp) : '0';
-
-    return this.ID_PREFIX + projectId;
-  },
   buildHTML: function(projectData) {
-    var id = this.getProjectCardId(projectData);
+    var id = projectData.id;
     var timestamp = new Date(projectData.Timestamp);
     var month = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"][timestamp.getMonth()];
     var dataForTemplate = {
@@ -24,6 +17,10 @@ var ProjectCard = {
       description: projectData.Description ? projectData.Description : "",
       thumbnail: projectData['Thumbnail URL'] ? projectData['Thumbnail URL'] : "",
       interest: projectData.Interest ? projectData.Interest : "",
+      tags: projectData.Tags,
+      issues: projectData.Issues,
+      program: projectData.Program,
+      type: projectData.Type,
       link: projectData.URL ? projectData.URL : "",
       getInvolvedLink: projectData['Get involved URL'] ? projectData['Get involved URL'] : "",
       detailViewLink: utility.getProjectDirectLink(id),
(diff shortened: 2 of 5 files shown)


index.html
```diff
@@ -91,7 +91,7 @@
         featured ? "featured" : "",
       ].join(" ");
     %>
-    <div id="<%= id %>" class="<%= classNames %>" data-id="<%= id %>" data-timestamp="<%= rawTimestamp %>">
+    <div id="<%- id %>" class="<%- classNames %>" data-id="<%- id %>">
```
mmmavis: just syntax fixing here `<%- foo %>` instead of `<%= foo %>`

js/project-card.js
```diff
@@ -1,16 +1,9 @@
 /* Project Card */
 
 var ProjectCard = {
-  ID_PREFIX: "p",
   template: _.template($("script#project-card-template").html()),
-  getProjectCardId: function(projectData) {
```
mmmavis: I moved thi method to `PulseMaker` in `pulse.js`. I think `id` should be part of the project `object`... not just as part of the project card template data.

js/pulse.js
```diff
@@ -197,12 +198,18 @@ var PulseMaker = {
...
+
+    return this.PROJECT_ID_PREFIX + projectId;
+  },
   'parseGoogleSheetData': function(driveData) {
     // Formats JSON data returned from Google Spreadsheet and formats it into
     // an array with a series of objects with key value pairs like "column-name":"value"
     var headings = {};
-    var newData = {};
-    var finalData = [];
+    var projects = {};
```
mmmavis: renaming vars

js/pulse.js
```diff
@@ -216,18 +223,20 @@ var PulseMaker = {
...
-        newData[row][headings[col]] = value;
+        projects[row][headings[col]] = value;
       }
     }
 
-    for(var k in newData){
-      finalData.push(newData[k]);
+    for(var rowNum in projects){
+      var projectData = projects[rowNum];
+      projectData.id = this.generatePulseProjectId(projectData);
```
mmmavis: changes in `parseGoogleSheetData` method are mostly var ranaming except this one - adding `id` property to project

index.html
acabunoc: Do we not need `data-timestamp` anymore?

mmmavis: yup, we never used that before. I think I added it just for testing and forgot to revert it back after.

js/pulse.js
```diff
@@ -197,12 +198,18 @@ var PulseMaker = {
   'toggleAdditionalFeatures' : function () {
     if (FEATURE.orphans) { typography.preventTextOrphans(); }
   },
+  generatePulseProjectId(projectData) {
+    var timestamp = projectData.Timestamp ? projectData.Timestamp : false;
+    var projectId = timestamp ? Date.parse(timestamp) : '0';
```
acabunoc: I know this was just copied from above, but I wanted to raise a couple things:

* Will all projects have a unique time stamp? What about batch import projects (we might just migrate all hive portfolio projects at once)?
* More of a design suggestion, but this isn't a very readable in the url. Can we use a slug version of the title + maybe append a timestamp? Something like what Medium does?

Neither point is a huge deal now, especially if the url structure will change down the road, but something I'd like to keep in mind.

acabunoc:
Changes look good to me! We might have ID issues down the road, but we can deal with that later

js/pulse.js
mmmavis: > Will all projects have a unique time stamp? What about batch import projects (we might just migrate all hive portfolio projects at once)?

@xmatthewx and I have been trying to figure out a way to generate unique ID for spreadsheet data (we have this issue with the event app too)... and we found timestamp being the most reliable/safe field to generate ID from. 

> More of a design suggestion, but this isn't a very readable in the url. Can we use a slug version of the title + maybe append a timestamp? Something like what Medium does?

I'm not sure how often project title gets updated, and if any changes happen in spreadsheet will it update the timestamp as well? @xmatthewx

mmmavis:
@acabunoc yea i agree! It's really tricky to generate unique and permanent ID for spreadsheet data, I've been struggling for so longgggggg. After all Google Spreadsheet it's not a real database 😜

PR #173, opened 2016-08-16
Cleaned up utility.js

### LINKED ISSUE #169: Tidy up `utility.js` by utilizing the Location API
Seems like I did a ton of redundant and complicated work in [`utility.js`](https://github.com/mozilla/network-pulse/blob/master/js/utility.js) when I could've made things a lot simpler and easier by using the [Location API](https://developer.mozilla.org/en-US/docs/Web/API/Location) 😁

mmmavis:
Fixes #169 

purely code cleanup by making use of the [Location API](https://developer.mozilla.org/en-US/docs/Web/API/Location)

@cadecairos @acabunoc review plz? 😄

### DIFF
--- a/js/utility.js
+++ b/js/utility.js
@@ -1,40 +1,29 @@
 /* utility */
 
 var utility = {
-  findContainer : function(elem, containerClass){
-      while ((elem = elem.parentElement) && !elem.classList.contains(containerClass));
-      return elem;
-  },
-  getProjectIdFromUrl: function(url) {
-    if ( url.indexOf("id=") > -1 ) {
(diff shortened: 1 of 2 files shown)


js/utility.js
```diff
@@ -1,40 +1,29 @@
 /* utility */
 
 var utility = {
-  findContainer : function(elem, containerClass){
```
mmmavis: old method that was never used

cadecairos:
Looks great, R+

mmmavis:
👯 👯 👯

PR #191, opened 2016-08-20
Ignored any project without timestamp

### LINKED ISSUE #148: Faved tab breaks if a faved project is later deleted from spreadsheet
I can fix this by filtering out those "no longer existed" projects before we render projects into the faved tab. 

Question: Do we also want to add some visual clue/notice so that users know their fav'ed project has been removed from the spreadsheet?

/cc @kristinashu @xmatthewx

mmmavis:
Fixes #148

Since we are generating `id` based on `timestamp`, any project without `timestamp` should be ignored.

/cc @cadecairos @acabunoc :

### DIFF
--- a/js/pulse.js
+++ b/js/pulse.js
@@ -139,10 +139,11 @@ var PulseMaker = {
     if (FEATURE.orphans) { typography.preventTextOrphans(); }
   },
   generatePulseProjectId: function(projectData) {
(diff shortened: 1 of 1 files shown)


acabunoc:
👍 looks good!

mmmavis:
❤️

PR #196, opened 2016-08-23
Included search terms as URL query string

### LINKED ISSUE #67: Search terms creates a unique url
As mentioned in https://github.com/mozilla/network-pulse/issues/58, if I enter a search term I should be able to copy that url and send it to someone so they go straight to a page with those search results (and any new ones that my have been posted during fitting the criteria). This is something that came up in an interview today and would be very useful.

@mmmavis what do you think?

mmmavis:
Fixes #67 

On Search mode...

 - [x] URL gets updated with search keyword when search bar loses focus
 - [x] URL gets updated with search keyword when search bar is in focus and user hits "enter"
 - [x] Grab the updated URL and visit it on a new tab. Make sure it brings you to the Search view with the keyword prefilled and only matching projects are shown.

/cc @cadecairos @acabunoc

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -7,6 +7,9 @@ var Search = {
     if (event.keyCode === 27) { // escape
       Search.deactivate();
     } else {
+      if ( event.keyCode === 13 ) { // enter
+         Search.updateUrlQuery(query);
+      }
       Search.showSearchResult(query);
     }
   },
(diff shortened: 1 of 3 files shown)


alicoding:
LGTM

mmmavis:
YAY @alicoding THANKS! 🍎 🍊 🍇 🍫 🍰

PR #208, opened 2016-09-12
Added mofo-bootstrap

### LINKED ISSUE #166: Use mofo-bootstrap
Not sure if we should use mofo-bootstrap or not? Building prototypes quickly seems like the perfect use of bootstrap but not sure we're there yet? But could be a good opportunity to test it?

Design elements we could pull from it: fonts, colors, buttons, maybe icons, maybe grid, maybe footer

Is it an issue that pulse is not built in react?

Thoughts? @xmatthewx @mmmavis @gvn

mmmavis:
Fixes #166


mmmavis:
Got r+ from @kristinashu . Merging this nowwwww ✨

PR #322, opened 2017-01-20
Logged in status note on favs page

### LINKED ISSUE #318: Add Sign in and Logout link on favs screen
Based on design in - Sign in and Logout #299 

screen shot 2017-01-19 at 10 50 53 am
screen shot 2017-01-19 at 10 51 58 am

mmmavis:
Fixes #318

@Pomax do you mind reviewing this one? 

Most changes are in `pages/favs.jsx ` and `config/config.js`. I'm relying on the `/nonce` call to verify a user's logged in state and to get his/her `username`... I'm not really making use of the `loggedin` param that's passed back from a `/login` API call since I have to make a `/nonce` call to get `username` anyway.


config/config.js
```diff
@@ -0,0 +1,30 @@
...
+
+let config = {
+  userLoggedInStatus: false,
+  username: ``
+};
+
+export default {
+  logInUrl: {
+    get: function(redirectUrl = env.ORIGIN) {
+      return `${env.PULSE_API}/login?original_url=${encodeURIComponent(redirectUrl)}`;
```
Pomax: Hm, why is redirectUrl not simply `window.location.toString()` here, so that users always get back to whatever page they're on right now?

config/config.js
```diff
@@ -0,0 +1,30 @@
+import env from "../config/env.generated.json";
+
+let config = {
```
Pomax: small nit, but since this isn't a configuration object calling it `config` is not being true to its use in the code. It's better to call this `userData` or something

config/config.js
```diff
@@ -0,0 +1,30 @@
...
+    get: function(redirectUrl = env.ORIGIN) {
+      return `${env.PULSE_API}/login?original_url=${encodeURIComponent(redirectUrl)}`;
+    }
+  },
+  userLoggedInStatus: {
+    get: function () {
+      return config.userLoggedInStatus;
+    },
+    set: function (status) {
+      config.userLoggedInStatus = ( status === `True` || status === true);
```
Pomax: `status.toString().toLowerCase() === 'true'` ?

js/build-env-config.js
```diff
@@ -13,6 +13,7 @@ if (environment !== `PRODUCTION`) {
 }
 
 var config = {
+  ORIGIN: habitat.get(`ORIGIN`),
```
Pomax: As there is no `.env` file, this feels like it needs local dev defaults. `ORIGIN: habitat.get('ORIGIN', 'http://localhost:3000')`?

pages/entry.jsx
```diff
@@ -15,10 +15,10 @@ export default React.createClass({
   fetchData(entryId = ``) {
     Service.entry
       .get(entryId)
-      .then((entry) => {
+      .then((response) => {
         this.setState({
           loadedFromGoogle: true,
-          entry: entry
+          entry: JSON.parse(response)
```
Pomax: this needs a try/catch, otherwise this call has the potential to completely crash the app. Ideally at this point the data is already correct though.

pages/favs.jsx
```diff
@@ -3,18 +3,61 @@ import { getFavs } from '../js/favs-manager';
...
     return {
-      favs: null
+      favs: null,
+      loggedInCheckTimestamp: null
     };
   },
+  verifyLoggedInStatus() {
+    Service.nonce()
+      .then((response) => {
+        this.updateLoggedInInfo(true, JSON.parse(response).user);
```
Pomax: this JSON.parse should probably happen first, (either in a try/catch or with a comment that this can throw), with its data then passed into this.updateLoggedInInfo, for clarity.

components/project-list/project-list.jsx
```diff
@@ -29,9 +29,9 @@ export default React.createClass({
   fetchData(params = {}) {
     Service.entries
       .get(params)
-      .then((entries) => {
+      .then((response) => {
         // sort data from mostly -> least recently added
-        entries = entries.sort((a,b) => {
+        let entries = JSON.parse(response).sort((a,b) => {
```
Pomax: Not so nice thing about JSON.parse: it'll throw, so this needs catching.

js/service.js
```diff
@@ -29,13 +29,14 @@ function doXHR(route, params = {}) {
   return new Promise((resolve, reject) => {
     request.open(`GET`, `${route}${combinedParams ? pojoToQuery(combinedParams) : ``}`, true);
 
+    request.withCredentials = true;
     request.onload = (event) => {
       let result = event.currentTarget;
 
       if (result.status >= 200 && result.status < 400) {
-        resolve(JSON.parse(result.response));
+        resolve(result.response);
```
Pomax: this is actually the right place to do the JSON parsing: either result.response is proper JSON in which case we send the parsed data on, or it's not and in that case this XHR handling code is responsible for saying what should happen on a failure because the data it received does not match the data it expected.

pages/favs.jsx
```diff
@@ -3,18 +3,61 @@ import { getFavs } from '../js/favs-manager';
...
+      })
+      .catch((reason) => {
+        this.updateLoggedInInfo(false);
+        console.error(reason);
+      });
+  },
+  updateLoggedInInfo(status, username = ``) {
+    config.userLoggedInStatus.set(status);
+    config.username.set(username);
+    this.setState({loggedInCheckTimestamp: new Date()});
```
Pomax: Your variable says "timestamp" but the content is a Date object. If this really is a timestamp, `Date.now()` - if it's a datetime, then this should be `loggedInCheckDateTime`

pages/favs.jsx
```diff
@@ -3,18 +3,61 @@ import { getFavs } from '../js/favs-manager';
...
+  },
   componentDidMount() {
+    // get IDs of user's favorited entries
     this.setState({favs: getFavs()});
+
+    // let's verify what the logged in status is
+    this.verifyLoggedInStatus();
   },
   render() {
+    let redirectBackUrl = `${env.ORIGIN}${this.props.router.getCurrentLocation().pathname}`;
```
Pomax: Is there a reason we can't just say `redirectBackUrl = window.location.toString();`?

pages/favs.jsx
```diff
@@ -3,18 +3,61 @@ import { getFavs } from '../js/favs-manager';
...
+    let logInUrl = config.logInUrl.get(redirectBackUrl);
+    let loggedInStatus = config.userLoggedInStatus.get() ?
+                        <p>Hi {config.username.get()}! <a href="" onClick={this.logOutUser}>Log out</a>.</p>
```
Pomax: Don't use an `<a>`  if you're not linking to an actual URL. This is not a hyperlinking navigation element, it's just a button, so let's use the correct markup: `<button tabIndex={0} onClick={..} className={'text-link-button'}>`with the classname imparting the correct styling to make the button look like clickable text in line with the rest of the client.

config/config.js
mmmavis: Agree. i had a hard time naming that object. Do you thinking it's fine to name this file `config.js`?

pages/favs.jsx
mmmavis: I just need `state` change to trigger component re-render. I'll change the var name to `loggedInCheckDateTime` for simplicity

pages/favs.jsx
mmmavis: Ah, `render` happened before component was mounted so I *assumed* we wouldn't have access to `window` in `render` until component was mounted. I just tested it and proved that I was wrong. Heh. Will update the PR.

pages/favs.jsx
mmmavis: hahahahah @Pomax you caught me! I've been using buttons for other places in the code but this one I was too lazy to style it initially... and eventually forgot to fix it before I submitted the PR.

config/config.js
mmmavis: ohhhhhhhh 😃

js/build-env-config.js
mmmavis: Hey @Pomax there's a default env file (`config/default.env`) where I put all the defaults there. It should load any missing env var from `default.env` if `.env` is not there.

```bash
ORIGIN='https://pulse-react.herokuapp.com'
PULSE_API='https://network-pulse-api-staging.herokuapp.com'
```

js/service.js
mmmavis: @Pomax in some cases we get redirected to the root route on Pulse API. It's status `200` but the response is just the html of the server's home page.

This is what I intend to change the above code to:

```js
if (result.status >= 200 && result.status < 400) {
  try {
    resolve(JSON.parse(result.response));
  } catch (error) {
    resolve(error);
  }
} else {
  reject(`XHR request failed, status ${result.status}.`);
}
```

config/config.js
Pomax: I'd probably rename this since it's not really setting up a `config`, it technically exports an API library, right? Maybe `userSettings.js`

js/build-env-config.js
Pomax: those defaults are not very useful when you're running localhost though?

mmmavis: Would adding a note to README to inform devs to create a `.env` file basd on `defaults.env` help? I'm worried if setting defaults with`habitat` plus loading potentially 2`.env` files might be a bit redundant?

config/config.js
mmmavis: There's server side generation going on as well so we can't simply set it to `window.location.toString()`

pages/favs.jsx
```diff
@@ -3,18 +3,64 @@ import { getFavs } from '../js/favs-manager';
...
   getInitialState() {
     return {
-      favs: null
+      favs: null,
+      loggedInCheckDateTime: null
     };
   },
+  verifyLoggedInStatus() {
+    Service.nonce()
+      .then((response) => {
```
Pomax: no need for parens in an arrow function if there's only one argument:
```
...
.then(response => this.updateLoggedInInfo(true, response.user))
...
```

Also no need for `{ return ... }` decoration if an arrow function is a one-liner, which is quite nice.

pages/favs.jsx
```diff
@@ -3,18 +3,64 @@ import { getFavs } from '../js/favs-manager';
...
+      })
+      .catch((reason) => {
+        this.updateLoggedInInfo(false);
+        console.error(reason);
+      });
+  },
+  updateLoggedInInfo(status, username = ``) {
+    config.userLoggedInStatus.set(status);
+    config.username.set(username);
+    this.setState({loggedInCheckDateTime: new Date()});
```
Pomax: `new Date()` is a Date object, which is almost certainly not what you want (for a string, you want to call `(new Date()).toLocaleString()` or the like, for a timestamp you want `Date.now()`), but more importantly: this state variable isn't used anywhere (not passed down, not used in rendering, etc) so let's remove it?

pages/favs.jsx
```diff
@@ -3,18 +3,64 @@ import { getFavs } from '../js/favs-manager';
...
 export default React.createClass({
   getInitialState() {
     return {
-      favs: null
+      favs: null,
+      loggedInCheckDateTime: null
     };
   },
+  verifyLoggedInStatus() {
+    Service.nonce()
```
Pomax: There is a new decidated `/userstatus/` route now, so we should probably switch this from `.nonce` to `.userStatus`

pages/favs.jsx
mmmavis: ahhhh so many new syntax to get used to. Thanks for pointing these out

components/project-list/project-list.jsx
```diff
@@ -29,14 +29,11 @@ export default React.createClass({
...
-        // sort data from mostly -> least recently added
-        entries = entries.sort((a,b) => {
-          return b.id - a.id;
-        });
+      .then((response) => {
+
         this.setState({
           loadedFromGoogle: true,
-          entries: entries
+          entries: response
```
Pomax: btw ES6 allows for a shorter object notation if the property (in-scope) has the same name as the value, so you could do:
```
.then(entries => this.setState({ loadedFromGoogle: true, entries });
```
And done: the `entries` part is treated as if you wrote `entries: entries`

pages/favs.jsx
mmmavis: I had `loggedInCheckDateTime` originally just to trigger `re-render` when it got updated. I will replace it with `this.forceUpdate()` instead.

js/service.js
```diff
@@ -17,25 +17,64 @@ function pojoToQuery(pojo) {
...
+/**
+ * Make an GET XHR request and return a promise to resolve it.
+ * Useful for making a call to an endpoint that doesn't return data.
+ * @param  {String} route A route fragment
+ * @param  {Object} params A POJO to be serialized as a query string
+ * @returns {Promise} A promise to resolve an XHR request
+ */
+function callURL(route, params = {}) {
+  let request = new XMLHttpRequest();
+  let combinedParams = Object.assign(params,defaultParams);
```
mmmavis: ohhhhhhhhhhhhhh thanks for catching this! 🍰

README.md
```diff
@@ -2,21 +2,9 @@
...
-
-### Production
-
-- `gh-pages` branch
-- https://mozilla.github.io/network-pulse
-
 ## 🔷 v2, React-based (upcoming)
 
-- https://pulse-react.herokuapp.com
+- https://pulse-react.herokuapp.com to check out what have been implemented
```
Pomax: what *has been

README.md
```diff
@@ -29,19 +17,24 @@ This is a prototype to test a few ideas on how to gather a stream of things crea
 
 - `git clone -b react https://github.com/mozilla/network-pulse.git`
 - `cd network-pulse`
+- `cp sample.env .env` (and modify values in sample.env so they match what you use for your local development env)
+    ```bash
+    HOST='the host you are using'
```
Pomax: I would recommend updating this so that it's explicit about what the values should be:

```bash
HOST=<url for the domain the pulse app runs on, without port suffix, and without trailing slash>
POST=<port number>
PULSE_API=<url for the pulse API server, without trailing slash>
```

README.md
```diff
@@ -29,19 +17,24 @@ This is a prototype to test a few ideas on how to gather a stream of things crea
...
+    PORT='the port number you are using'
+    PULSE_API='URL to your local Pulse API instance. Remove this var if you don't have one set up.'
+    ```
 - `npm install`
 - `npm start`
 
+To set up a local instance of Pulse API, follow instructions on https://github.com/mozilla/network-pulse-api/blob/master/README.md.
```
Pomax: *follow the instructions on

config/config.js
```diff
@@ -0,0 +1,33 @@
+import env from "../config/env.generated.json";
```
Pomax: I'd still recommend renaming this to `user-data.js`, because the thing this file exports is not a configuration object at all.

config/default.env
```diff
@@ -1 +1,2 @@
+HOST='https://pulse-react.herokuapp.com'
```
Pomax: this should not need quotes. Same for the pulse_api var

js/service.js
```diff
@@ -8,7 +8,7 @@ let defaultParams = {
 };
 
 /**
- * Serialize a POJO as a URL query string fragment
+ * Serialize a POJO(Plain Old JavaScript Object. A set of key-value pairs.) as a URL query string fragment
```
Pomax: this is such a bad name... A JavaScript object is arbitrarily complex, whereas this function expects a flat object. I would strongly recommend we rename this to `toQueryArguments()` or something, and rewrite this a little so it doesn't blow up in our face when you *do* pass in something complicated. 

Right now, this code assumes that `input[somekey]` serializes to a string just fine, which is quite untrue. While clever, this code has some issues:
```
  return Object.keys(pojo).reduce((previousValue, currentValue, index) => {
    return `${previousValue}${index !== 0 ? `&` : ``}${currentValue}=${pojo[currentValue]}`;
  }, `?`);
```
The two main issues here are that

1. If a key has value `null` or `undefined`, it should not end up in the argument list, but this code adds it anyway (but `false` *should* end up in query args), and
2. there is no test for `${pojo[currentValue]}` to see if it makes sense, so this code will happy generate `[Object object]` without any indication that something just went horribly wrong.

So I'd recommend we rewrite this to something that does what we need with a bit more control and clarity:
```
function toQueryPair(key, data) {
  let val = data[key];
  if (val === undefined || val === null) {
    return false;
  }
  let type = typeof val;
  switch(type) {
    case "object": return false;
    case "function": return false;
    default: return `${key}=${val.toString()}`
  }
}

function toQueryString(data) {
  const pairs = Object.keys(data).map( key => toQueryPair(key, data)).filter(e => !!e);
  const queryArguments = pairs.join('&');
  return `?${queryArguments}`;
}
```
This code is certainly longer, but at every step we control whether or not data that should (or should not) be allowed is converted, explicitly, to string form and packed as URL argument key=value pair.

js/service.js
```diff
@@ -19,25 +19,64 @@ function pojoToQuery(pojo) {
...
+ * Useful for making a call to an endpoint that does return data.
  * @param  {String} route A route fragment
  * @param  {Object} params A POJO to be serialized as a query string
  * @returns {Promise} A promise to resolve an XHR request
  */
-function doXHR(route, params = {}) {
+function getDataFromURL(route, params = {}) {
   let request = new XMLHttpRequest();
-  let combinedParams = Object.assign(params,defaultParams);
+  let combinedParams = Object.assign(defaultParams,params);
```
Pomax: `defaultParams, params`? (space after comma)

js/service.js
Pomax: this looks like hasn't been updated?

js/service.js
```diff
@@ -19,25 +19,64 @@ function pojoToQuery(pojo) {
...
+ * @param  {String} route A route fragment
+ * @param  {Object} params A POJO to be serialized as a query string
+ * @returns {Promise} A promise to resolve an XHR request
+ */
+function callURL(route, params = {}) {
+  let request = new XMLHttpRequest();
+  let combinedParams = Object.assign(params,defaultParams);
+
+  return new Promise((resolve, reject) => {
+    request.open(`GET`, `${route}${combinedParams ? pojoToQuery(combinedParams) : ``}`, true);
```
Pomax: let's use `toQueryString`

pages/favs.jsx
```diff
@@ -1,20 +1,69 @@
...
+        console.error(reason);
+      });
+  },
   componentDidMount() {
+    let query = this.props.router.location.query;
+    
+    // we don't need the "loggedin" param sent back from Pulse API to be presented at all times
+    delete query.loggedin;
+
+    browserHistory.push({
```
Pomax: you want a `replace` here, rather than a `push`, so that the browser "forgets" there was ever a `loggedin` query argument to begin with.

sample.env
```diff
@@ -0,0 +1,3 @@
+HOST='http://localhost'
```
Pomax: these values should not need quotes

js/service.js
mmmavis: Changes have been made. It's just this review interface is confusing 👀

mmmavis: i take that back. i did forget to update this one heh

package.json
```diff
@@ -7,7 +7,7 @@
     "bootstrap": "shx rm -rf dist && shx mkdir dist",
     "start": "npm run build && run-p server watch:**",
     "build": "run-s bootstrap build:**",
-    "build:config": "node js/build-config.js > config/env.generated.json",
+    "build:config": "node js/build-env-config.js > config/env.generated.json",
```
Pomax: it looks like PORT is missing from `default.env` resulting in an `env.generated.json` with `ORIGIN:"https://....:undefined"`

js/build-env-config.js
```diff
@@ -12,10 +12,16 @@ if (environment !== `PRODUCTION`) {
   habitat.load(`config/default.env`);
 }
 
-var config = {
+const HOST = habitat.get(`HOST`);
+const PORT = habitat.get(`PORT`);
+const ORIGIN = HOST + (PORT ? `:${PORT}` : ``);
+const CONFIG = {
```
Pomax: ES6 for the win:
```
const CONFIG = { HOST, PORT, ORIGIN, PULSE_API: habitat.get(`PULSE_API`)};
```
This is the same as what you wrote, but ES6 improved object notation allows you to use `x` for any object propert:value pair `x:x`. Super convenient.

mmmavis:
yay yay yay 🍾

PR #334, opened 2017-01-24
log in/out flow

### LINKED ISSUE #309: Build "add new entry" form
Based on design in https://github.com/mozilla/network-pulse/issues/257

Comps: https://redpen.io/p/nzcd35d57dd04ad6dc

Basic styling is based on Bootstrap 4 http://v4-alpha.getbootstrap.com/content/reboot/#forms
but errors are styled a little different.

mmmavis:
Related to #309 

- log in/out flow
- /add page CSS


components/navbar/navbar.jsx
```diff
@@ -3,17 +3,6 @@ import { IndexLink, browserHistory } from 'react-router';
 import SelectorLink from '../selector-link/selector-link.jsx';
 
 export default React.createClass({
-  handleAddPageLinkClick(event) {
```
mmmavis: this was needed when /add page was just an embedded Google Form

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+  },
+  handleCreatorClick(event) {
+    event.preventDefault();
+    this.setState({numCreatorFields: this.state.numCreatorFields+1});
+  },
   render() {
+    let creatorFields = [];
+
+    for (var i=0; i<this.state.numCreatorFields; i++) {
+      creatorFields.push( <CreatorInputField key={i} />);
```
mmmavis: not sure what other `key` we can assign these fields with

mmmavis:
hey @Pomax can you help review this one? Especially the login/out flow part? 😄 

I was hoping to check user's login status (and send user to /login API route if he/she is not logged in) during React Router Route's `onEnter` but it seems like we are only allowed to modify a `Location` object there so redirecting user to an external URL is not possible....

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+      loggedInStatusChecked: false
+    };
+  },
+  componentWillMount() {
+    console.log(`UserData.userLoggedInStatus.get() = `, UserData.userLoggedInStatus.get());
+  },
+  componentDidMount() {
+    let loggedInParam = this.props.router.location.query.loggedin;
+
+    if ( !loggedInParam || loggedInParam.toString().toLowerCase() === `true` ) {
```
Pomax: This code (and the verify and update functions) should live in their own file that you can require in, and then call as
```
const userStatus = require('./userstatus')
...
componentDidMount() {
  if (this.props.router.location.query) {
    userstatus.handle(this.props.router.location.query);
  }
}
```
And then because `Service` is shared, anything `userstatus` sets pertaining to the user status will be in effect inside this class, too.

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+      // if there's no "loggedin" param presented in the URL OR if "loggedin" param is true
+      // we want to verify if user is really logged in by making a /userstatus call to Pulse API
+      this.verifyLoggedInStatus();
```
Pomax: This leads to the wrong codepath. Right now with this code in place, the form page will do a user login check, and when the user isn't logged in says "Sign in failed" which is wrong: no sign in even occurred. Instead when there is no `loggedin` query argyment, this code should says "Please log in to post a new entry" or some such, with the same link we present on the `favs` page.

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+      </div>
+    );
+
     return (
       <div className="add-page">
-        <div id="add-project-form">
-          <iframe src="https://docs.google.com/forms/d/1CiNYA3gzHPd4HYrgQuQB2KM0TDpsbn6jbMcl-0_OjM4/viewform?embedded=true" width="100%" height="100%" frameBorder="0" marginHeight="0" marginWidth="0">Loading...</iframe>
-        </div>
+        {
+          this.state.loggedInStatusChecked ? ( UserData.userLoggedInStatus.get() ? ContentForAuthedUser : ContentForUnauthedUser )
```
Pomax: if you need nested ternaries, don't use ternaries. Instead write out the code in an easier to read and maintain format.

pages/add/add.jsx
Pomax: This is actually a symptom of an architectural flaw: either this is a pure HTML form, in which case this component should be stateless, or this is a data UI, in which case the component should track all the form data in its state, and the form should have `onChange` handling for each field so that things like dynamically adding/removing creator fields with proper keys is actually possibe.

Right now you've written the form in a way that hands off all the functionality to the browser instead of React, and so if the component needs to rerender for whatever reason, the form as the user has it filled in so far will get emptied. So, if we want to be able to add more creators for something with more than three creators, this would need to be rewritten.

So the main question is: how much work do we want to do now, vs. later? We can present a static form with three creator fields, and then we should hard code that, and remove any kind of state from this component: it doesn't have one. 

Or, if we want a variable number of creators, then we need to rewrite this to be a proper React form based off of state for all the fields we care about, in which case we want to use a form builder for that (we have one over on https://github.com/mozilla/learning.mozilla.org/blob/master/pages/clubs/form/builder/Form.jsx but there might be others on npm these days, too).

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+      // we want to verify if user is really logged in by making a /userstatus call to Pulse API
+      this.verifyLoggedInStatus();
+    } else if ( loggedInParam && loggedInParam.toString().toLowerCase() === `false` ) {
+      // if "loggedin" param is presented in the URL and it is set to false
+      // we assume that user cannot be logged in using his/her Google Account
+      // (i.e., not using a Mozilla Staff account)
+      this.updateLoggedInInfo(false);
+    }
+  },
+  verifyLoggedInStatus() {
```
Pomax: this function show not exist here, this is "no related to the component in the slightest" functionality and so should just be in its own lib

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+
+          window.location.href = UserData.logInUrl.get(redirectUrl);
+        }
+      })
+      .catch(reason => {
+        this.updateLoggedInInfo(false);
+        console.error(reason);
+      });
+  },
+  updateLoggedInInfo(status, username = ``) {
```
Pomax: this function should not exist here, this is "not related to the component in the slightest" functionality and so should just be in its own lib

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+    UserData.userLoggedInStatus.set(status);
+    UserData.username.set(username);
+
+    // we have verified user's logged in status with Pulse API
+    this.setState({
```
Pomax: This should not be necessary.

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+      loggedInStatusChecked: true
+    });
+
+    // After the login process user will be redirected back to Pulse.
+    // There will be a query argument "loggedin" presented in the URL
+    // which is something we don't want it to stay at all times.
+    this.clearLoggedinParam();
+  },
+  clearLoggedinParam() {
```
Pomax: this function should not exist here, this is "not related to the component in the slightest" functionality and so should just be in its own lib

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+
+    // we don't need the "loggedin" param sent back from Pulse API to be presented at all times
+    delete query.loggedin;
+
+    browserHistory.replace({
+      pathname: this.props.router.location.pathname,
+      query: query
+    });
+  },
+  logOutUser(event) {
```
Pomax: this function should not exist here, this is "not related to the component in the slightest" functionality and so should just be in its own lib

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+        browserHistory.push({
+          pathname: `/featured`
+        });
+      })
+      .catch(reason => {
+        console.error(reason);
+        // TODO:FIXME: what do we do here?
+      });
+  },
+  handleCreatorClick(event) {
```
Pomax: We need to change how this works, and we should discuss this in mattermost

Pomax:
Architecture wise, here's what I suggest:

- consult UserData to see if the user is logged in. If not, show some UI that prompts the user to log in, similar to the `favs` page. 
- move all the code that is involved in "checking the user's login state" to its own file, with a simple API that lets you throw the current location at it and it will do everything that needs to happen based on `loggedin` query parameters.
- if we want an "uncontrolled" form, handling purely offloaded to the browser, then this code mostly works, but it needs a bit of cleanup.
- if we want a "controlled" form, we'll need to rewrite this form to using `onChange` handlers and have the class use its state to track form values. This is, admittedly, a lot more work.

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
 import React from 'react';
+import { browserHistory } from 'react-router';
+import HintMessage from '../../components/hint-message/hint-message.jsx';
+import Service from '../../js/service.js';
+import UserData from '../../js/user-data.js';
+import env from "../../config/env.generated.json";
+
+const CreatorInputField = () => {
+  return (
+    <input name="creators" type="text" placeholder="Name" width="150" className="form-control" />
```
Pomax: Right now these are inline blocks, so clicking "add another creator" does not add a new creator as a line, it adds it as if it's text and things look a little funky

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+                                  </HintMessage>);
+
+    let ContentForAuthedUser = (
+      <div>
+        <h1>Share with the Network</h1>
+        <p>Do you have something to share? If it might be useful to someone in our network, share it here! Pulse includes links to products and software tools, research reports and findings, think pieces, white papers, interviews, and curricula. If it might be useful, share it … at any stage or fidelity.</p>
+        <form action="" method="post">
+          <h2>Basic Info</h2>
+          <div className="posted-by">
+            <p>Posted by:&nbsp;<span className="user-full-name">{UserData.username.get()}</span></p>
```
Pomax: "Posting as" - the posting has not happened yet, so "posted by" is a little odd. Also we don't need the `&nbsp;` here, you only need that if the `<span>` is on the next line (because whitespace at the end of lines gets ignored.

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+                                               header={`Sign in failed`}
+                                               btn={{to: `/featured`, text: `Explore featured`}}>
+                                    <p>Only Mozilla staff can login now as we test this new platform. Check back soon!</p>
+                                  </HintMessage>);
+
+    let ContentForAuthedUser = (
+      <div>
+        <h1>Share with the Network</h1>
+        <p>Do you have something to share? If it might be useful to someone in our network, share it here! Pulse includes links to products and software tools, research reports and findings, think pieces, white papers, interviews, and curricula. If it might be useful, share it … at any stage or fidelity.</p>
+        <form action="" method="post">
```
Pomax: this is missing the URL to post to.

pages/add/add.jsx
```diff
@@ -1,12 +1,201 @@
...
+          <fieldset hidden>
+            <label className="form-control-label">
+              Posted by:&nbsp;
+              <input name="" type="text" className="form-control" value="" readOnly />
+            </label>
+            <p>Not you? <a href="">Sign out.</a></p>
+          </fieldset>
+          <label className="form-control-label">
+            Title of the project<em className="required">(required)</em>
+            <input name="title" type="text" placeholder="Title" className="form-control" />
```
Pomax: if this is required, then this should have `required={"required"}`, so that the form can't be submitted until it has a value.

mmmavis:
@Pomax I moved all the login logics to a separate file `js/login.js`. Am I on the right track? (note I haven't touched any form bits in the commit I just pushed)

assets/svg/icon-user.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
Pomax: good idea to run this through svgo, there's quite a bit of stuff in this SVG that doesn't need to be there (Lines 3, 4, 5, 6 are all unnecessary, the element `id` attributes do nothing, `.00000` does nothing, etc.)

components/hint-message/hint-message.jsx
```diff
@@ -8,7 +8,8 @@ const HintMessage = (props) => {
         <h2><img src={props.imgSrc} /></h2>
         <h2>{props.header}</h2>
         {props.children}
-        <Link to={props.btn.to} className="btn">{props.btn.text}</Link>
+        { props.btn ? <Link to={props.btn.to} className="btn">{props.btn.text}</Link>
```
Pomax: kinda worried about the use of `btn` as variable name. Can we name this after what we use it for? Because this code suggests it should be called `internalLink`, controlling whether something is a react-router link or html link.

js/login.js
```diff
@@ -0,0 +1,68 @@
+import { browserHistory } from 'react-router';
+import Service from './service.js';
+import UserData from './user-data.js';
+
+export default {
+  clearLoggedInParam(locationObj) {
+    let query = locationObj.query;
+
+    delete query.loggedin;
```
Pomax: Just because `delete` is a really expensive operation, let's wrap this as
```
if (query.loggedin) {
  delete query.loggedin;

  browserHistory.replace({
    pathname: locationObj.pathname,
    query: query
  });
}
```
While there *should* be a `loggedin` query param, we can never be 100% certain. The internet has been known to break.

js/login.js
```diff
@@ -0,0 +1,68 @@
...
+      // After the login process user will be redirected back to Pulse.
+      // There will be a query argument "loggedin" presented in the URL
+      // which is something we don't want it to stay at all times.
+      this.clearLoggedInParam(currentLocation);
+    }
+
+    // verify user's logged in status with Pulse API
+    Service.userstatus()
+      .then(response => {
+        let type = `unknown`;
```
Pomax: For v2, we don't need these checks. Someone is either staff (or at least someone we trust enough to give a moz ldap account), or they're an anonymous user right now.

js/login.js
```diff
@@ -0,0 +1,68 @@
...
+    let query = locationObj.query;
+
+    delete query.loggedin;
+
+    browserHistory.replace({
+      pathname: locationObj.pathname,
+      query: query
+    });
+  },
+  verifyLoggedInStatus(currentLocation, callback) {
```
Pomax: can we rename `callback` to have a name that actually indicates when it's getting called back, like `onUserStatusFinalised` or something? This makes understanding how the code work without first reading the entire file much easier. Or, given that we either call it with nothing or an error, `finishUserUpdate`

js/login.js
```diff
@@ -0,0 +1,68 @@
...
+        callback();
+      })
+      .catch(reason => {
+        console.error(reason);
+        this.updateUserInfo(`unknown`);
+        callback(reason);
+      });
+  },
+  updateUserInfo(type, username = ``) {
+    UserData.type.set(type);
```
Pomax: if you use ES6 classes with `get` and `set`, you don't call those in your "real" code, you just use the variable as if it was a normal class property. So in this case you simply call `UserData.type = type` and whatever code you specified for `set type(value) {....}` will automatically run.

js/user-data.js
```diff
@@ -2,6 +2,7 @@ import env from "../config/env.generated.json";
 
 let userData = {
   loggedInStatus: false,
+  type: `unknown`, // three possible values: staff, nonstaff, unknown
```
Pomax: this seems unnecessary for v2, let's leave it off until v3?

pages/add/add.jsx
```diff
@@ -14,85 +14,28 @@ const CreatorInputField = () => {
...
-    UserData.userLoggedInStatus.set(status);
-    UserData.username.set(username);
+    let currentLocation = this.props.router.location;
 
-    // we have verified user's logged in status with Pulse API
-    this.setState({
-      loggedInStatusChecked: true
+    Login.verifyLoggedInStatus(currentLocation, error => {
+      if (!error) {
+        this.forceUpdate();
```
Pomax: why the forceupdate?

pages/add/add.jsx
```diff
@@ -14,85 +14,28 @@ const CreatorInputField = () => {
...
     event.preventDefault();
-    Service.logout()
-      .then(() => {
-        // we have successfully logged user out
-        this.updateLoggedInInfo(false);
-        // redirect user to /featured page
+
+    Login.logoutUser(error => {
+      if (!error) {
         browserHistory.push({
```
Pomax: does this just fake the URL or actually redirect?

pages/add/add.jsx
```diff
@@ -105,28 +48,35 @@ export default React.createClass({
...
+
+
+    let contentForNonStaff= ( <HintMessage imgSrc={`/assets/svg/icon-user.svg`}
                                                header={`Sign in failed`}
                                                btn={{to: `/featured`, text: `Explore featured`}}>
                                     <p>Only Mozilla staff can login now as we test this new platform. Check back soon!</p>
                                   </HintMessage>);
 
-    let ContentForAuthedUser = (
+    let contentForStaff = (
```
Pomax: This should just be content, not specifically "For staff". If people are logged in, they get to see this content.

pages/add/add.jsx
```diff
@@ -190,12 +140,21 @@ export default React.createClass({
       </div>
     );
 
+
+    let Content;
```
Pomax: we only need to toggle on logged in vs. not logged in, not on staff vs non-staff vs. anonymous

pages/bookmarks.jsx
```diff
@@ -14,57 +13,45 @@ export default React.createClass({
...
-      .then(() => {
-        // we have successfully logged user out
-        this.updateLoggedInInfo(false);
-      })
-      .catch(reason => {
-        console.error(reason);
-      });
+    Login.logoutUser(error => {
+      if (!error) {
+        this.forceUpdate();
```
Pomax: is this to make sure the page is reloaded in a state based on "now an anonymous user"?

pages/bookmarks.jsx
```diff
@@ -14,57 +13,45 @@ export default React.createClass({
...
+    });
   },
   render() {
     let redirectUrl = `${env.ORIGIN}${this.props.router.getCurrentLocation().pathname}`;
     let logInUrl = UserData.logInUrl.get(redirectUrl);
-    let loggedInStatus = UserData.userLoggedInStatus.get() ?
-                        <p>Hi {UserData.username.get()}! <button onClick={this.logOutUser} className="btn-link">Log out</button>.</p>
-                        : <p>Are you Mozilla Staff? <a href={logInUrl}>Sign in</a>.</p>;
-    
+    let promptToLogIn = <p>Are you Mozilla Staff? <a href={logInUrl}>Sign in</a>.</p>;
```
Pomax: again, too many types here? Let's just focus on "logged in" vs. "anonymous" for now.

Pomax:
A gist of how to "unify" the login and user functionality (not a true unification, they're still two objects but only the User object gets to call Login helper functions, nothing else should) over on https://gist.github.com/Pomax/13984de64dc597b21fbd849f6e8bacc6 with a comment on how this can then be used on a page over on https://gist.github.com/Pomax/13984de64dc597b21fbd849f6e8bacc6#gistcomment-1982466

components/hint-message/hint-message.jsx
```diff
@@ -8,8 +8,8 @@ const HintMessage = (props) => {
         <h2><img src={props.imgSrc} /></h2>
         <h2>{props.header}</h2>
         {props.children}
-        { props.btn ? <Link to={props.btn.to} className="btn">{props.btn.text}</Link>
-                    : <a href={props.link.href} className="btn">{props.link.text}</a> }
+        { props.internalLink && props.linkText ? <Link to={props.internalLink} className="btn" onClick={props.onClick}>{props.linkText}</Link> : null }
```
Pomax: we can probably make this less repeatigin, but let's do that in follow-up

pages/add/add.jsx
```diff
@@ -140,21 +141,21 @@ export default React.createClass({
       </div>
     );
 
+    let getContent = function() {
```
Pomax: This should not be an inline function, but just a normal class function that returns function calls for the various bits of code.
```
  getContent() {
      if (user.loggedin) {
        return this.contentForLoggedInUser();
      }
      if (user.failedLogin) {
        return failurePrompt();
      }
      return anonymousContent();
    };
```
And then those three functions can, nice and contained, generate the content only when necessary, rather than generating all the content first, and then throwing away what we won't be using.

pages/add/add.jsx
```diff
@@ -140,21 +141,21 @@ export default React.createClass({
...
-    } else {
-      content = promptToLogIn;
-    }
+      return anonymousContent;
+    };
 
     return (
       <div className="add-page">
-        { content }
+        { getContent() }
```
Pomax: along with the above note, `this.getContent()`

pages/bookmarks.jsx
```diff
@@ -3,54 +3,50 @@ import { getBookmarks } from '../js/bookmarks-manager';
...
     return {
-      bookmarks: null
+      bookmarks: null,
+      user
     };
   },
+  handleSignInBtnClick(event) {
+    event.preventDefault();
+
+    let redirectUrl = `${env.ORIGIN}${this.props.router.getCurrentLocation().pathname}`;
```
Pomax: Hmm, doesn't this code only work in a browser, when a user clicks a button? It kind of feels like `user.login()` should know to use `window.location.toString()` as its redirect URL, unless there is no `window` variable available... thoughts?

pages/bookmarks.jsx
```diff
@@ -3,54 +3,50 @@ import { getBookmarks } from '../js/bookmarks-manager';
...
-    let redirectUrl = `${env.ORIGIN}${this.props.router.getCurrentLocation().pathname}`;
-    let logInUrl = UserData.logInUrl.get(redirectUrl);
-    let promptToLogIn = <p>Are you Mozilla Staff? <a href={logInUrl}>Sign in</a>.</p>;
-    let contentForStaff = <p>Hi {UserData.username.get()}! <button onClick={this.handleLogOutBtnClick} className="btn-link">Log out</button>.</p>;
-    let contentForNonStaff = <p>Only Mozilla staff can login now as we test this new platform. Meanwhile, your favorited projects can still be remembered on this device.</p>;
+    let anonymousContent = <p>Are you Mozilla Staff? <button className="btn btn-link" onClick={this.handleSignInBtnClick}>Sign in</button>.</p>;
+    let contentForLoggedInUser = <p>Hi {user.username}! <button onClick={this.handleLogOutBtnClick} className="btn-link">Log out</button>.</p>;
+    let failurePrompt = <p>Only Mozilla staff can login now as we test this new platform. Meanwhile, your favorited projects can still be remembered on this device.</p>;
+
+    let getloggedInStatus = function() {
```
Pomax: don't inline functions in `render`, just make them class functions and invoke them as `this.getLoggedInStatus()` etc. Especially in this case where `user` is global rather than local to the class instance.

assets/svg/icon-user.svg
Pomax: Looking at this SVG closer, it seems the only thing it gives us is a large FontAwesome icon... is there a reason we need this as SVG image? If so, we probably want to vectorise the glyph rather than relying on the FontAwesome font. If not, we already have FontAwesome loaded so we could just show this in HTML with a `<span>` without any need for an SVG file

pages/add/add.jsx
```diff
@@ -1,12 +1,161 @@
...
+          </div>
+          <fieldset hidden>
+            <label className="form-control-label">
+              Posted by:&nbsp;
+              <input name="" type="text" className="form-control" value="" readOnly />
+            </label>
+          </fieldset>
+          <label className="form-control-label">
+            Title of the project<em className="required">(required)</em>
+            <input name="title" type="text" placeholder="Title" className="form-control" />
```
Pomax: This is still missing a `required="required"` attribute, to make sure title *must* be filled in.

Pomax: This is still missing a `required="required"` attribute, to make sure title *must* be filled in.

pages/add/add.jsx
```diff
@@ -1,12 +1,161 @@
...
+                                      internalLink={`/featured`}
+                                      linkText={`Explore featured`}>
+                          <p>Only Mozilla staff can login now as we test this new platform. Check back soon!</p>
+                        </HintMessage>);
+
+    let contentForLoggedInUser = (
+      <div>
+        <h1>Share with the Network</h1>
+        <p>Do you have something to share? If it might be useful to someone in our network, share it here! Pulse includes links to products and software tools, research reports and findings, think pieces, white papers, interviews, and curricula. If it might be useful, share it … at any stage or fidelity.</p>
+        <form action="" method="post">
```
Pomax: this form still needs to point to an actual place. Or, we need to *not* use a `submit` button and make sure to set a `preventDefault` on the form itself, because we're going to have to do a manual submission to the API server.

pages/add/add.jsx
```diff
@@ -1,12 +1,161 @@
...
+              <input name="" type="text" className="form-control" value="" readOnly />
+            </label>
+          </fieldset>
+          <label className="form-control-label">
+            Title of the project<em className="required">(required)</em>
+            <input name="title" type="text" placeholder="Title" className="form-control" />
+          </label>
+          <label className="form-control-label">
+            URL<em className="required">(required)</em>
+            <input name="" type="url" placeholder="https://example.com" className="form-control" />
```
Pomax: This `<input>` is still missing a `required="required"` attribute, to make sure the content URL *must* be filled in.

pages/add/add.jsx
```diff
@@ -1,12 +1,161 @@
...
+          </label>
+          <label className="form-control-label">
+            Describe what you are sharing. Keep it simple and use plain language.
+            <textarea name="description" placeholder="Description" className="form-control" />
+          </label>
+          <h2>Optional Details</h2>
+          <fieldset className="creators">
+            <legend>Who are the creators? This could be staff, contributors, partners…</legend>
+            { creatorFields }
+            <button onClick={this.handleCreatorClick} className="btn btn-link">&#43; Add another creator</button>
```
Pomax: note that the creatorFields still add "new fields" inline instead of as blocks:
screenshot 152

pages/add/add.jsx
mmmavis: Oh yea the form part is definitely not finished yet. I wanted to get login logics working first. Maybe I should remove the form chunks from this PR and create followup PR to take care it separately?

Pomax: Ah, yeah that's probably better.

pages/add/add.jsx
```diff
@@ -1,12 +1,84 @@
...
+      pathname: `/featured`
+    });
+  },
+  handleCreatorClick(event) {
+    event.preventDefault();
+    this.setState({numCreatorFields: this.state.numCreatorFields+1});
+  },
+  getContentForLoggedInUser() {
+    return( <div>
+              <p>(TODO:FIXME: form for logged in user to be built)</p>
```
Pomax: ❤️

mmmavis:
🍾 🎉  thanks SO MUCH for your patience and guidance @Pomax 👍 👍 👍 👍 👍

Pomax:
quiet welcome! 🍰

PR #339, opened 2017-01-25
Make sure we are passing the right params props to <ProjectList>

### LINKED ISSUE #333: "featured" seems to poll the same list as "favs"
If I load the client, fave three entries, and visit `/favs` I see those three entires. Clicking back to `featured` however then still only shows those three entires. It seems that `featured` is using the same list that `favs` is using?

mmmavis:
Fixes #333

### DIFF
--- a/js/service.js
+++ b/js/service.js
@@ -50,7 +50,8 @@ function toQueryString(data) {
  */
 function getDataFromURL(route, params = {}) {
   let request = new XMLHttpRequest();
-  let combinedParams = Object.assign(defaultParams, params);
+  let clonedDefaultParams = Object.assign({}, defaultParams); // so we don't mutate defaultParams itself in the next line
+  let combinedParams = Object.assign(clonedDefaultParams, params);
 
   return new Promise((resolve, reject) => {
     request.open(`GET`, `${route}${combinedParams ? toQueryString(combinedParams) : ``}`, true);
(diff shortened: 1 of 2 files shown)


js/service.js
```diff
@@ -50,7 +50,8 @@ function toQueryString(data) {
  */
 function getDataFromURL(route, params = {}) {
   let request = new XMLHttpRequest();
-  let combinedParams = Object.assign(defaultParams, params);
+  let clonedDefaultParams = Object.assign({}, defaultParams); // so we don't mutate defaultParams itself in the next line
+  let combinedParams = Object.assign(clonedDefaultParams, params);
```
Pomax: No need to split this over three lines, Object.assign thankfully takes "as many objects as necessary":
```
let combinedParams = Object.assign({}, defaultParams, params);
```

js/service.js
```diff
@@ -87,7 +88,8 @@ function getDataFromURL(route, params = {}) {
  */
 function callURL(route, params = {}) {
   let request = new XMLHttpRequest();
-  let combinedParams = Object.assign(defaultParams, params);
+  let clonedDefaultParams = Object.assign({}, defaultParams); // so we don't mutate defaultParams itself in the next line
```
Pomax: same here, we can just pass a new `{}` as first argument

Issue #340, opened 2017-01-26
fix up Object.assign


Pomax:
https://github.com/mozilla/network-pulse/pull/339 landed before review, and should get cleaned up as per  the comments in https://github.com/mozilla/network-pulse/pull/339/files/b0e45e8c828ca253dfca6238787c24b9fc1e0604

closes #333 when this gets closed.

mmmavis:
PR was approved and merged!

PR #341, opened 2017-01-27
Object.assign improvements

### LINKED ISSUE #333: "featured" seems to poll the same list as "favs"

mmmavis:
Fixes #333, #340


Pomax:
r+

PR #345, opened 2017-01-30
Used browerlist package instead of passing options to cli command directly

### LINKED ISSUE #323: "postcss" node package does not work on Windows
We should try to find an alternative or at least have a note on README first.

mmmavis:
Fixes #323 

@Pomax Do you mind testing on Windows? I've never developed on Windows machine before... I'm just worried it would take me a while to have the basics set up in Windows VM. 🙄

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,13 +11,13 @@
     "build:client": "run-s build-client:**",
     "build-client:js": "webpack --config webpack.config.js --optimize-minimize --display-error-details",
     "build-client:scss": "node-sass scss/main.scss dist/css/main.css",
-    "build-client:prefix:main": "postcss --use autoprefixer --autoprefixer.browsers 'last 3 versions' -o dist/css/main.css dist/css/main.css",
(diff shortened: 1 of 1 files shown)


Pomax:
O_O postcss no longer breaks!

mmmavis:
YEAH 💃 🎉 🍾  :man_dancing: 

and thanks for sharing the links. Hopefully installing `node` and all other things won't be too different from what it's like on Mac 😁

Pomax:
@gvn @mavis this looks like it's worth updating the mofo-* docs/boilerplate code with, so that future postcss use won't break npm scripts for Windows users

mmmavis:
@gvn @Pomax remind me what those repos are. Is it just https://github.com/MozillaFoundation/mofo-style ?

Pomax:
probably more https://github.com/MozillaFoundation/mofo-standards in the form of a section that explains that if autoprefixing is required, we use `postcss` with `browserslist` (and perhaps even explicitly say that we do *not* use `postcss` with the standard `autoprefixer` package because that doesn't work for Windows users)

PR #347, opened 2017-01-31
Show splash screen on user's first visit only

### LINKED ISSUE #331: show the splash page on first use, suppress on subsequent loads
As the splash page comes with a forced waiting period, it'll be a good idea to show it on first-time-use, but then suppress it on subsequent loads by using a localStorage variable like `suppressSplashScreen=true`.

If the localStorage config does not contain the var, or it's not `true`, show the splash screen, and set `suppressSplashScreen=true` at the same time, so that a next load will bypass the splash.

mmmavis:
Fixes #331 

@Pomax 👀 ?

### DIFF
--- a/routes.jsx
+++ b/routes.jsx
@@ -38,6 +38,7 @@ const App = React.createClass({
       // wait for CSS animation to finish first before we
       // set `suppressSplashScreen` in localStorage to true
       localstorage.setItem(`suppressSplashScreen`, `true`);
+      this.setState({ suppressSplashScreen: true });
     });
   },
   renderWelcomeSplash() {
--- a/routes.jsx
+++ b/routes.jsx
@@ -36,7 +36,8 @@ const App = React.createClass({
     document.querySelector(`#app`).classList.add(`splash-dismissed`);
     this.refs.splash.addEventListener(`transitionend`, () => {
       // wait for CSS animation to finish first before we
-      // set `suppressSplashScreen` in localStorage to true
+      // set `suppressSplashScreen` in localStorage and 
+      // this.state.suppressSplashScreen to true
       localstorage.setItem(`suppressSplashScreen`, `true`);
       this.setState({ suppressSplashScreen: true });
     });
--- a/routes.jsx
+++ b/routes.jsx
@@ -27,21 +27,25 @@ const App = React.createClass({
     };
   },
   componentDidMount() {
-    if (!this.state.suppressSplashScreen) {
+    if (this.refs.splash) {
       setTimeout(this.dismissSplash, this.props.dismissTimeout);
     }
   },
   dismissSplash() {
-    this.splash.classList.add(`dismissed`);
+    this.refs.splash.classList.add(`dismissed`);
     document.querySelector(`#app`).classList.add(`splash-dismissed`);
-    localstorage.setItem(`suppressSplashScreen`, `true`);
+    this.refs.splash.addEventListener(`transitionend`, () => {
+      // wait for CSS animation to finish first before we
+      // set `suppressSplashScreen` in localStorage to true
+      localstorage.setItem(`suppressSplashScreen`, `true`);
+    });
   },
   renderWelcomeSplash() {
     if (this.state.suppressSplashScreen) {
       return null;
     }
     return (
-      <div id="splash" ref={(splash) => { this.splash = splash; }}>
+      <div id="splash" ref="splash">
         <div className="container">
           <h1><img src="/assets/svg/pulse-wordmark.svg" width="200" height="46" alt="Mozilla Network Pulse" /></h1>
           <p>A stream of assets from peers across the Mozilla Network.</p>
--- a/routes.jsx
+++ b/routes.jsx
@@ -18,17 +18,23 @@ import Footer from './components/footer/footer.jsx';
 const App = React.createClass({
   getInitialState() {
     return {
-      suppressSplashScreen: localstorage.getItem(`suppressSplashScreen`) === `true`
+      suppressSplashScreen: (localstorage.getItem(`suppressSplashScreen`) === `true`)
+    };
+  },
+  getDefaultProps() {
+    return {
+      dismissTimeout: 3000
     };
   },
   componentDidMount() {
-    if (this.state.suppressSplashScreen !== `true`) {
-      setTimeout(function() {
-        this.splash.classList.add(`dismissed`);
-        document.querySelector(`#app`).classList.add(`splash-dismissed`);
-      }, 3000);
-      localstorage.setItem(`suppressSplashScreen`, `true`);
+    if (!this.state.suppressSplashScreen) {
+      setTimeout(this.dismissSplash, this.props.dismissTimeout);
     }
+  },
+  dismissSplash() {
+    this.splash.classList.add(`dismissed`);
+    document.querySelector(`#app`).classList.add(`splash-dismissed`);
+    localstorage.setItem(`suppressSplashScreen`, `true`);
   },
   renderWelcomeSplash() {
     if (this.state.suppressSplashScreen) {
@@ -44,7 +50,6 @@ const App = React.createClass({
     );
   },
   render() {
-
     return (
       <div>
         { this.renderWelcomeSplash() }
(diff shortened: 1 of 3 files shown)


routes.jsx
```diff
@@ -15,21 +15,38 @@ import Navbar from './components/navbar/navbar.jsx';
...
+    };
+  },
   componentDidMount() {
-    setTimeout(function() {
-      this.splash.classList.add(`dismissed`);
-      document.querySelector(`#app`).classList.add(`splash-dismissed`);
-    }, 3000);
+    const suppressSplashScreenKey = `suppressSplashScreen`;
+
+    this.setState({
```
Pomax: instead, put this in `getInitialState`

routes.jsx
```diff
@@ -15,21 +15,38 @@ import Navbar from './components/navbar/navbar.jsx';
 import Footer from './components/footer/footer.jsx';
 
 const App = React.createClass({
+  getInitialState() {
+    return {
```
Pomax: the code that is currently in componentDidMount should be here, so that the `suppressSplashScreen` value is correct immediately (remember, localStorage exists well before the app does, so don't need to "wait" for the app, the value exists irrespective of the app)

routes.jsx
```diff
@@ -15,21 +15,38 @@ import Navbar from './components/navbar/navbar.jsx';
...
+
     return (
       <div>
-        <div id="splash" ref={(splash) => { this.splash = splash; }}>
-          <div className="container">
-            <h1><img src="/assets/svg/pulse-wordmark.svg" width="200" alt="Mozilla Network Pulse" /></h1>
-            <p>A stream of assets from peers across the Mozilla Network.</p>
-          </div>
-        </div>
+        { !this.state.suppressSplashScreen ?
```
Pomax: I would put this in its own function. So something like
```
{ this.renderWelcomeSplash() }
```
And then have `renderWelcomeSplash` be
```
...
renderWelcomeSplash() {
  if (this.state.suppressSplashScreen) {
    return null;
  }
  return (
    <div ...>
      ...
    </div>
  );
},
...
```

routes.jsx
mmmavis: I'm getting server side error `ReferenceError: localStorage is not defined` if I try to access `localStorage` before `componentDidMount`. Should I shim it as what you did in `app-user.js` ?

mmmavis: i.e., create a lib file that has the shim code, and instead of just doing `localStorage.method()` I do `require('storage').method()`

mmmavis:
@Pomax PR updated!

routes.jsx
```diff
@@ -15,21 +16,38 @@ import Navbar from './components/navbar/navbar.jsx';
...
+    return {
+      suppressSplashScreen: localstorage.getItem(`suppressSplashScreen`) === `true`
+    };
+  },
   componentDidMount() {
-    setTimeout(function() {
-      this.splash.classList.add(`dismissed`);
-      document.querySelector(`#app`).classList.add(`splash-dismissed`);
-    }, 3000);
+    if (this.state.suppressSplashScreen !== `true`) {
```
Pomax: `this.state.suppressSpashScreen` is a boolean, not a string.

routes.jsx
```diff
@@ -15,21 +16,38 @@ import Navbar from './components/navbar/navbar.jsx';
 import Footer from './components/footer/footer.jsx';
 
 const App = React.createClass({
+  getInitialState() {
+    return {
+      suppressSplashScreen: localstorage.getItem(`suppressSplashScreen`) === `true`
```
Pomax: To make it more obvious that this is a boolean, not a string, it might be worth putting this in parentheses.

routes.jsx
```diff
@@ -15,21 +16,38 @@ import Navbar from './components/navbar/navbar.jsx';
...
-    setTimeout(function() {
-      this.splash.classList.add(`dismissed`);
-      document.querySelector(`#app`).classList.add(`splash-dismissed`);
-    }, 3000);
+    if (this.state.suppressSplashScreen !== `true`) {
+      setTimeout(function() {
+        this.splash.classList.add(`dismissed`);
+        document.querySelector(`#app`).classList.add(`splash-dismissed`);
+      }, 3000);
+      localstorage.setItem(`suppressSplashScreen`, `true`);
```
Pomax: I'd move this line into the above function, so we make sure to skip the splash after the user has seen the splash for the full 3 seconds at least once. And mostly because we're working in React, I'm also going to say "let's not define functions on the fly. This can be a `dismissSplash` function on the React component itself, which we call with `setTimeout(this.dismissSplash, this.props.dismissTimeout)` (with a `defaultProps` that sets `dismissTimeout: 3000`)

routes.jsx
```diff
@@ -15,21 +16,38 @@ import Navbar from './components/navbar/navbar.jsx';
...
+    }
+  },
+  renderWelcomeSplash() {
+    if (this.state.suppressSplashScreen) {
+      return null;
+    }
+    return (
+      <div id="splash" ref={(splash) => { this.splash = splash; }}>
```
Pomax: I don't understand this line. Why the need to assign this to an instance variable instead of just calling it `ref="splash"` and then accessing it using `this.refs.splash`? That's what refs are for. As you're only referencing it after mounting, there should be no problems relying on the normal ref behaviour.

mmmavis: @Pomax I'm following React's preferred pattern for `ref`

https://facebook.github.io/react/docs/refs-and-the-dom.html
> Using the `ref` callback just to set a property on the class is a common pattern for accessing DOM elements. The preferred way is to set the property in the ref callback like in the above example. There is even a shorter way to write it: `ref={input => this.textInput = input}`.

routes.jsx
mmmavis: @Pomax do you mind explaining more about 

> because we're working in React, I'm also going to say "let's not define functions on the fly".

I'd love to know why it's not a good idea to do that in React 😉

routes.jsx
Pomax: that is a really weird thing to recommend. I can't say I agree with that at all: that DOM node has no business becoming permanently aliased at the component level, especially as it won't get cleaned up: when the splash render function returns `null`, what is `this.splash`? As such, I'd strongly recommend staying with `this.refs` instead:
```
if (this.refs.splash) {
  // assign class so it fades away
}
```
because there might not even be a splash screen to work with =)

routes.jsx
```diff
@@ -15,21 +16,43 @@ import Navbar from './components/navbar/navbar.jsx';
...
-    setTimeout(function() {
-      this.splash.classList.add(`dismissed`);
-      document.querySelector(`#app`).classList.add(`splash-dismissed`);
-    }, 3000);
+    if (!this.state.suppressSplashScreen) {
+      setTimeout(this.dismissSplash, this.props.dismissTimeout);
+    }
+  },
+  dismissSplash() {
+    this.splash.classList.add(`dismissed`);
```
Pomax: is there a way for us to tie a `setState({ suppressSplashScreen: true })` to when this CSS transition finishes?

Pomax:
I don't like bypassing `this.refs` so the code is technically fine but I'd really like to not use that `this.splash` binding because the vast majority of the application's runtime that value will be `null` due to the fact that the splash screen never even got generated after the first time.

mmmavis:
PR updated!

The [`transitioned` event](https://developer.mozilla.org/en-US/docs/Web/Events/transitionend) seems to be well-supported so I attached an handler to make sure we call ```localstorage.setItem(`suppressSplashScreen`, `true`);``` after the fading animation finishes. For users on browsers that don't support the`transitioned` event, worst case they will see the splash screen every time they visit Pulse...

routes.jsx
```diff
@@ -27,21 +27,25 @@ const App = React.createClass({
...
   },
   dismissSplash() {
-    this.splash.classList.add(`dismissed`);
+    this.refs.splash.classList.add(`dismissed`);
     document.querySelector(`#app`).classList.add(`splash-dismissed`);
-    localstorage.setItem(`suppressSplashScreen`, `true`);
+    this.refs.splash.addEventListener(`transitionend`, () => {
+      // wait for CSS animation to finish first before we
+      // set `suppressSplashScreen` in localStorage to true
+      localstorage.setItem(`suppressSplashScreen`, `true`);
```
Pomax: not just localstorage, also in state, so let's also add a `this.setState({ suppressSplashScreen: true })` here to make sure render() `null`s the splash div

mmmavis:
PR updated

mmmavis:
🎉

PR #348, opened 2017-02-04
Search page tech improvement. Uses Pulse API to get search result.

### LINKED ISSUE #328: Search page tech
### UX wise 
Current plan: We would like to have "search as you type". And it should search through all entry fields/meta (@xmatthewx @kristinashu correct me if I'm wrong)

### Tech wise
Before we had pagination on Pulse API, I used to just get all entries by one `/entries` call and bind search input box's `keyUp()` event with client side filtering. The client side filter searches through *all* entry meta (not just `title` or `description`) and returns entries that have meta matched the search keyword.
Since now pagination on Pulse API is enabled by default, I wonder what the best way is to implement the search functionality.

- **Challenge 1**: multiple `/entries` calls have to be made in order to fetch all entries in the database
- **Challenge 2**: can we make use the `/entries?search=` param (note it currently searches through title and description only)? If so, does that mean we have to make a new API call on every keystroke? 😨 

Maybe we should set some limitation on how search works. e.g., don't enable search-as-you-type. instead let 'Enter' trigger the search?

/cc @Pomax @alanmoo @kristinashu @xmatthewx

mmmavis:
Fixes #328


mmmavis:
@alanmoo since you've done similar things on the Science site, can you review this PR? 🍻

components/project-list/project-list.jsx
```diff
@@ -66,15 +88,6 @@ export default React.createClass({
       return entries.sort((a,b) => {
         return bookmarkedIdArray.indexOf(a.id.toString()) > bookmarkedIdArray.indexOf(b.id.toString());
```
mmmavis: Is it worth doing this sorting on Pulse API side instead?

e.g, `/entries/?ids=3,1,5` will return entry no.3 first, followed by no.1, and lastly no.5.

alanmoo:
First thing, before I forget: you need to add `shx` to `package.json`

components/project-list/project-list.jsx
alanmoo: Is this something you're using in a lot of places? The db returns these in numerical order by default, unfortunately so it's non-trivial to make it behave like this. That's not to say it's incredibly difficult, I'm just not sure it's worth  the lift here if there's a JS function that does what you need relatively easily.

mmmavis: If that's the case don't worry about it. We just need to do the sorting once - for the initial /fav (bookmark) page render. :bowtie:

pages/search/search.jsx
```diff
@@ -17,32 +18,29 @@ export default React.createClass({
...
-    this.updateSearchQuery();
+    this.refs.searchInput.setState({value: ``}, () => {
+      this.updateSearchQuery();
+    });
   },
-  handleInputKeyUp(event) {
-    if (event.keyCode === 27) { // escape key
-      this.clearSearch();
-    }
+  handleInputChange() {
```
alanmoo: If this doesn't do anything else, can you just call `updateSearchQuery` from the Debounce input itself?

components/project-list/project-list.jsx
```diff
@@ -16,28 +16,50 @@ export default React.createClass({
...
+        ids: null
+      }
+    };
+  },
   componentDidMount() {
     this.fetchData(this.props.params);
   },
   componentWillReceiveProps(nextProps) {
+    // Reset this.state.entries
+    this.resetEntries();
```
alanmoo: I don't think this should be called here. If I type in something new, it shouldn't clear the results while it's doing the search to find the new set; Doing that makes it feel broken because you see a moment of "0 results found", which isn't true. It's especially confusing if you're refining your search and the results you had disappear, and then the same ones reappear.

components/project-list/project-list.jsx
```diff
@@ -16,28 +16,48 @@ export default React.createClass({
...
+        ids: null
+      }
+    };
+  },
   componentDidMount() {
     this.fetchData(this.props.params);
   },
   componentWillReceiveProps(nextProps) {
+    // Reset state
+    this.resetState();
```
mmmavis: We should reset state as soon as a new search has been initiated. However, instead of showing "0 results found" I've update that message to `Searching for entries that match ‘some_keyword’...` (see `render()` below)

components/project-list/project-list.jsx
```diff
@@ -88,9 +99,14 @@ export default React.createClass({
...
-    if (this.props.params && this.props.params.search && projects) {
-      // show search result if on search page and search keyword is presented
-      searchResult = (<p>{projects.length} {projects.length > 1 ? `results` : `result`} found for ‘{this.props.params.search}’</p>);
+    if (this.props.params && this.props.params.search) {
+      if (projects) {
+        // Show search result
+        searchResult = (<p>{projects.length} {projects.length > 1 ? `results` : `result`} found for ‘{this.props.params.search}’</p>);
+      } else {
+        // We are still waiting to hear back from Pulse API. Let's show some 'searching' notice in the meantime.
+        searchResult = (<p>Searching for entries that match ‘{this.props.params.search}’...</p>);
```
mmmavis: @kristinashu @alanmoo 
`Searching for entries that match ‘some_keyword’...` <= plz help fixing this sentence that I came up with 😂

components/project-list/project-list.jsx
alanmoo: But most of the time you won't even have a chance to read that text if the connection is reasonably fast- this process results in a "flicker" of an intermediate state, instead of just changing from one set of results to another. If you want something in the middle, might I suggest a spinner or something to indicate something's happening in the background, without immediately clearing the results?

Think of the case where someone searches for "web" and then decides to add "literacy" to the end of it- it doesn't make sense that the results disappear before reappearing- they should just reduce to entries with the complete string "web literacy" in them.

mmmavis: Since we are doing search-as-you-type (we initiate the search when there's no activity happening in the input box for x amount of time), for me having the previous search result stayed while new search is happening in the background can confuse me. If there's no UI change go along with it, as an user I won't know a new search has already been initiated. Having a spinner instead of text `Searching for entries that match ‘some_keyword’...` sounds great! But I do think we should clear previous search result while fetching results from Pulse API for the current search keyword.

@alanmoo @kristinashu thoughts? This is what has been implemented in this PR
pulse search 2

alanmoo: But imagine the case where a user is typing just slow enough to trigger multiple searches- they'll have content appearing and disappearing seemingly at random.

kristinashu: I don't see the problem. I don't think we should reset the state as soon as a user start to type (it should continue to show results for "web" as you add "literacy"). It's ok to have a slight delay. Also don't think we even need a message or a spinner. Message is definitely too quick to read, so maybe a spinner but let's try having nothing and see if users complain.

mmmavis: > But imagine the case where a user is typing just slow enough to trigger multiple searches- they'll have content appearing and disappearing seemingly at random.

Um I guess that's the nature of search-as-you-type. We don't start new search on every keystroke so to me that's fair enough. For non slower typers having previous search result stayed can make it look like the system is having problem detecting the new keyword user just typed, especially when network connection is bad.

Anyways, since you two both agree on the same approach I'll update my PR 😄

alanmoo: That's why I suggest a spinner or something to indicate that it's doing something, just not a message you have to read with disappearing content.

components/project-list/project-list.jsx
xmatthewx: How about just "Searching for..."

components/project-list/project-list.jsx
xmatthewx: Spinners are nice. 

I haven't read this whole thread, but if we get stuck on this, I suggest we consider adding a _Go!_ button. Find as you type is only more awesome if it feels more awesome.

mmmavis: @alanmoo 

> That's why I suggest a spinner or something to indicate that it's doing something, just not a message you have to read with disappearing content.

I'm a bit confused about this part. Did you mean in the transition state (user searched for 'web' first and then 'web literacy') we show

```
input box: [ web literacy ]

spinner icon

[ matching result #1 for 'web' ] [ result #2 ] [ result #3 ]
[ matching result #4 for 'web' ] [ result #6 ] [ result #6 ]
...
...

```

mmmavis:
@alanmoo PR updated. Didn't add spinner but we can always add one later if needed.

alanmoo:
I spun this up and accidentally (but in a reproducible way) got it into an inconsistent state. I clicked search, then typed `w e space backspace b`.

image

Looking at my network activity, the search for `web` was triggered and responded to (it has 73 items on staging), but this was never reflected in the UI count. When trying to reproduce it, I sometimes see the status flicker from `73 results found for web' to `119 results found for we `. It seems then like it's a race condition, where the `web` search returns first, but then the `we ` search returns and takes priority. I suspect the issue is in the promise starting at line 68 of `search.jsx`. The result with more pages is taking longer to fetch, so it's setting state last.

mmmavis:
i think i know what went wrong, will update!

mmmavis:
this should do... (hopefully)

pages/search/search.jsx
```diff
@@ -51,24 +50,97 @@ export default React.createClass({
...
+                          ref="searchInput"
+                          onChange={this.handleInputChange}
+                          placeholder="Search keywords, people, tags..." />
           <button className="btn dismiss" onClick={this.handleDismissBtnClick}>&times;</button>
         </div>
-        <ProjectList params={{search: this.state.searchQuery}} />
+        <div className="project-list">
+          { searchResult }
+          { projects ? <div className="projects">{projects}</div> : null }
+          { showViewMoreBtn ? <div className="view-more"><button type="button" className="btn" onClick={this.handleViewMoreClick}>View more</button></div> : null }
```
alanmoo: This isn't hooked up

pages/search/search.jsx
```diff
@@ -51,24 +50,97 @@ export default React.createClass({
...
+    params.page = apiPageIndex;
+
+    Service.entries
+      .get(params)
+      .then((data) => {
+        entries = entries.concat(data.results);
 
+        if (data.next) {
+          // there are more matched entries in the database we need to fetch
+          this.fetchData(params, entries, apiPageIndex+1);
```
alanmoo: Why do we need to fetch all the pages up front if we've got a load more button later on?

pages/search/search.jsx
```diff
@@ -1,48 +1,47 @@
...
-      this.clearSearch();
+  searchByQueryInUrl(searchQueryInUrl) {
+    let inputBoxValue = this.refs.searchInput.state.value;
+
+    if (searchQueryInUrl !== inputBoxValue) {
+      // make sure input box value is updated with searchQueryInUrl
+      // then use searchQueryInUrl as the search param to fetch data from Pulse API
+      this.refs.searchInput.setState({value: searchQueryInUrl ? decodeURIComponent(searchQueryInUrl) : ``});
+      this.setState({
+        keywordBeingSearched: searchQueryInUrl
```
alanmoo: What's the purpose of this state variable? It seems like it's only checked on line 85, which could just read the value of the input box directly, no?

pages/search/search.jsx
```diff
@@ -51,24 +50,97 @@ export default React.createClass({
...
+  handleDismissBtnClick() {
+    this.clearSearch();
   },
   render() {
+    let projects;
+    let showViewMoreBtn;
+    let searchResult;
+
+    if (this.state.keywordSearched) {
+      let numEntriesMatched = this.state.entriesMatched.length;
```
alanmoo: If we don't fetch all the entries at the start, we can just use the `count` property on the API response here

components/project-list/project-list.jsx
```diff
@@ -11,47 +8,55 @@ export default React.createClass({
...
-        entries: []
-      });
+    // <ProjectList> doesn't always get unmounted and remounted when navigating between pages.
+    // (e.g., When navigating between the /issue/issue-name pages <Issue><ProjectList></Issue>
+    //        do not get re-mounted since the same components are being rendered.)
+    // It is treated as passing new props to <ProjectList {...newProps} />
+    let newIssueEntered = (nextProps.params.issue !== this.props.params.issue);
 
-      // fetch data based on the new params props
+    if ( newIssueEntered ) {
```
alanmoo: Why issues specifically? And where does the issue prop actually get used? This makes me feel like something's not wired up to work best with React. Do you need a higher level component to fetch the data and then pass the data set down to the project list?

pages/search/search.jsx
mmmavis: oh yea that's better, will update.

pages/search/search.jsx
mmmavis: @alanmoo 

1) I didn't realize the `count` property returned from the API is the total count of all matched entries, not just how many entries are in that particular `page`. I thought I had to add them all up to get the total of matched entries count. Can we update the property name to something not so generic? Or at least have some docs for it?

2) Not sure how big our db will grow but for since we only have ~200 entries now, making one fetch is enough to retrieve all matched entries (`page_size` is set to 996, the closest number to 1000 that's divisible by 6). Currently every time "Load more" is clicked we display 24 more entries. Users see those entries immediately after they click the button. We *could* change it so that we don't fetch all data at once when we load the page, we make new Pulse API call every time users click on "Load more". This means users will have to wait till we hear back from the call. If we decide to make this change I suggest bumping up " the number of entries in each batch" from 24 to a bigger number. Otherwise it can be quite annoying as users have to go through "click and wait" cycle for many times just to load a few hundreds of entries. /cc @kristinashu @xmatthewx I would like to know your thoughts on this 😉

components/project-list/project-list.jsx
mmmavis: from `routes.jsx`

```jsx
<Route path="issues">
  <IndexRoute component={Issues} />
  <Route path=":issue" component={Issue} />
</Route>
```

And my inline comments in `project-list.jsx`
```jsx
// <ProjectList> doesn't always get unmounted and remounted when navigating between pages.
// (e.g., When navigating between the /issue/issue-name pages <Issue><ProjectList></Issue>
//        do not get re-mounted since the same components are being rendered.)
// It is treated as passing new props to <ProjectList {...newProps} />
```

And `issue` prop is being used in the next line after the one we are commenting here on
```jsx
if ( newIssueEntered ) {
  // Fetch data based on the new params props to ensure data gets fetched and displayed accordingly.
  this.fetchData(nextProps.params);
}
```

mmmavis: ^ is the reasoning why I implemented it that way. Better solution is welcome!

pages/search/search.jsx
xmatthewx: >  If we decide to make this change I suggest bumping up " the number of entries in each batch" from 24 to a bigger number. 

👍 as long as we don't overwhelm mobile users, loading 48 or more is fine by me. we could also pre-fetch as user scrolls down and nears the bottom. 

With page_size of 996, are we fetching that many _results_ or searching that many projects in total?  In 2017, we might cross 1000 entries to _search_. We would need to lift that limit b/c pace will increase as people anticipate new network site. I wouldn't, however, expect to cross 1000 search _results_ this year (for meaningful searches), so a cap on results is fine.

pages/search/search.jsx
mmmavis: nice catch!

components/project-list/project-list.jsx
mmmavis: oh yea but we should tackle it in a separate ticket/PR.

pages/search/search.jsx
Pomax: @mmmavis a lot of paginating APIs use `count` for the total count, with `page` and `per_page` for you to figure out exactly which set in `[0...count]` you're looking at, so I'd recommend we keep it like that.

alanmoo: @mmmavis 

1. This is standard API behavior and the docs at the top of [the entries endpoint](https://network-pulse-api-staging.herokuapp.com/entries/) already document the `page_size`, so it doesn't make sense to change this 

    `?page_size=` - Number of results on a page. Defaults to 48

2. Yeah I guess I question the purpose of the "Load more" button at all. Only thing I can think of is to avoid forcing the browser to load all those images unless they're wanted. Maybe we do fetch them all at once, and the "load more" button is a gate-keeper which then opens up to show the rest of the results at once? Presumably

components/project-list/project-list.jsx
alanmoo: Ah, I see what's going on, but agreed it should be a separate ticket. I'll file it.

scss/main.scss
```diff
@@ -169,10 +166,25 @@ button[type="submit"] {
...
+.project-list {
+  .projects {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+    margin-left: -$project-card-side-margin;
+    margin-right: -$project-card-side-margin;
+  }
+
+  .view-more {
```
alanmoo: Can you use a `.text-center` class on the element instead?

scss/main.scss
```diff
@@ -169,10 +166,25 @@ button[type="submit"] {
   color: $form-text-light;
 }
 
+// project-list
+
+.project-list {
+  .projects {
```
alanmoo: If you use `row flex-items-center` as the classes instead, it does the same thing (with a minor difference in margin), and then you don't need this CSS. (By consistently using the bootstrap row/column classes, you avoid the margin issue.)

alanmoo:
Refactor suggestions detailed in https://github.com/mozilla/network-pulse/issues/361

scss/main.scss
mmmavis: cool will file followup PR

mmmavis:
@alanmoo thanks I'll tackle #361 soon. Just wanna get this one in place first 😂 

and @alanmoo I can't find our comments around "load more" button behaviour anymore... where did it go?!?!? 

(*dislike* GitHub's new review interface)

PR #353, opened 2017-02-09
Use helmet

### LINKED ISSUE #350: Use helmet
Use https://www.npmjs.com/package/helmet to secure our Express with various HTTP headers.

mmmavis:
Fixes #350 

@cadecairos @Pomax Is it as simple as just adding `app.use(helmet());` to `server.js` ? I thought using the [defaults](https://www.npmjs.com/package/helmet#how-it-works) should be fine but since I have limited knowledge on HTTP headers I'd like you to help verify these configs are correctly set. Thanks!

### DIFF
--- a/js/security-headers.js
+++ b/js/security-headers.js
@@ -7,7 +7,8 @@ export default {
     ],
     scriptSrc: [
       `'self'`,
-      `'unsafe-inline'`
+      `'unsafe-inline'`,
+      `https://*.google-analytics.com`
     ],
     fontSrc: [
       `'self'`,
@@ -21,7 +22,8 @@ export default {
     imgSrc: [
       `'self'`,
       `data:`,
-      `*`
+      `https:`,
+      `http:`
     ],
     connectSrc: [
       `'self'`,
--- a/js/security-headers.js
+++ b/js/security-headers.js
@@ -1,0 +1,33 @@
+import env from "../config/env.generated.json";
+
+export default {
+  directives: {
+    defaultSrc: [
+      `'none'`
+    ],
+    scriptSrc: [
+      `'self'`,
+      `'unsafe-inline'`
+    ],
+    fontSrc: [
+      `'self'`,
+      `https://code.cdn.mozilla.net`
+    ],
+    styleSrc: [
+      `'self'`,
+      `'unsafe-inline'`,
+      `https://code.cdn.mozilla.net`
+    ],
+    imgSrc: [
+      `'self'`,
+      `data:`,
+      `*`
+    ],
+    connectSrc: [
+      `'self'`,
+      env.PULSE_API || `https://network-pulse-api-staging.herokuapp.com/`
+    ]
+  },
+  reportOnly: false,
+  browserSniff: false
+};
--- a/server.js
+++ b/server.js
@@ -20,9 +20,7 @@ import env from "./config/env.generated.json";
 const defaultPort = 3000;
 const PORT = env.PORT || process.env.PORT || defaultPort;
 
-/**
- * Some app security settings
- */
+// Some app security settings
 
 app.use(helmet.contentSecurityPolicy(securityHeaders));
 
--- a/server.js
+++ b/server.js
@@ -1,5 +1,6 @@
 import express from 'express';
 import path from 'path';
+import helmet from 'helmet';
 import React from 'react';
 import { renderToString } from 'react-dom/server';
 import { match, RouterContext } from 'react-router';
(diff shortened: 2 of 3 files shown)


Pomax:
We do a fair bit of helmet security in learning, over on https://github.com/mozilla/learning.mozilla.org/blob/master/app.js#L54-L76

server.js
```diff
@@ -18,8 +19,66 @@ import env from "./config/env.generated.json";
...
+      `\'self\'`,
+      env.PULSE_API || `https://network-pulse-api-staging.herokuapp.com/`,
+      `*`
+    ]
+  },
+  reportOnly: false,
+  browserSniff: false
+};
+
+app.use(helmet.contentSecurityPolicy(securityHeaders));
```
cadecairos: Nice to see a CSP header, we should consider doing more work later to remove the unsafe directives we're adding (unsafe-inline, data:, "*")

server.js
```diff
@@ -18,8 +19,66 @@ import env from "./config/env.generated.json";
...
 const PORT = env.PORT || process.env.PORT || defaultPort;
 
+/**
+ * Some app security settings
+ */
+
+var securityHeaders = {
+  directives: {
+    defaultSrc: [
+      `\'self\'`
```
cadecairos: could this be 'none' without breaking anything?

server.js
mmmavis: All of the project thumbnails we are showing (e.g., https://pulse-react.herokuapp.com/featured) are hotlink images. That's why I had to add ``*`` to `imgSrc`. 

We also have anchors that link out to each of the project's homepage so ``*`` is there in `connectSrc`. (though after reading docs on MDN I'm still not super sure about what I was supposed to put for `connectSrc`). We make XHR to Pulse API and have external links to places that we can't predict. I guess in that case it makes sense that I `*` there???

server.js
```diff
@@ -18,8 +19,66 @@ import env from "./config/env.generated.json";
...
+    ],
+    styleSrc: [
+      `\'self\'`,
+      `\'unsafe-inline\'`,
+      `https://code.cdn.mozilla.net`
+    ],
+    imgSrc: [
+      `\'self\'`,
+      `\'unsafe-inline\'`,
+      `data:`,
```
mmmavis: The reason why I put `data:` here is that the footer styling we are going to pull from `mofo-bootstrap` encodes SVG in stylesheet. That's why I *thought* we should allow `data:` here?

server.js
cadecairos: connectSrc only applies to XHR requests, so we can drop the wildcard.

mmmavis: kk will update. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src the "<a> ping" confused me. 😉

server.js
```diff
@@ -18,8 +19,65 @@ import env from "./config/env.generated.json";
 const defaultPort = 3000;
 const PORT = env.PORT || process.env.PORT || defaultPort;
 
+/**
+ * Some app security settings
+ */
+
+var securityHeaders = {
```
Pomax: totally worth putting this in its own `security-headers.js` or something.

server.js
```diff
@@ -18,8 +19,65 @@ import env from "./config/env.generated.json";
...
+  directives: {
+    defaultSrc: [
+      `\'none\'`
```
Pomax: `\'` is nothing unless the string itself is quoted with single quotes, so if we're using templating strings there should be no reason whatsoever to escape the single quotes inside of that. Do these need single quotes at all?

server.js
Pomax: that is correct reasoning =)

server.js
```diff
@@ -18,8 +19,65 @@ import env from "./config/env.generated.json";
...
+    ]
+  },
+  reportOnly: false,
+  browserSniff: false
+};
+
+app.use(helmet.contentSecurityPolicy(securityHeaders));
+
+app.use(helmet.xssFilter({
+  setOnOldIE: true
```
Pomax: Microsoft no longer supports old IE, as of January 2016 it only officially supports IE11 and EDGE, so I'm pretty okay taking this out. If you're on an old version of IE, our website is the least of your problems. @cadecairos opinions?

server.js
```diff
@@ -18,8 +19,65 @@ import env from "./config/env.generated.json";
...
+}));
+
+// maxAge for HSTS header must be at least 18 weeks (see https://hstspreload.org/)
+app.use(helmet.hsts({
+  maxAge: 60 * 60 * 24 * 7 * 18 // 18 weeks in seconds
+}));
+
+app.use(helmet.ieNoOpen());
```
Pomax: same reasoning as above for this option.

server.js
cadecairos: I don't think there's harm in leaving it in.

server.js
mmmavis: I'll remove `/`s. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src) single quotes are required so I'll leave those there.

mmmavis:
@Pomax 👀 ?

server.js
```diff
@@ -18,8 +20,32 @@ import env from "./config/env.generated.json";
 const defaultPort = 3000;
 const PORT = env.PORT || process.env.PORT || defaultPort;
 
+/**
+ * Some app security settings
+ */
+
```
Pomax: this is function comment syntax (at least as per jsdoc/javadoc) so you probably want this to be `// Some app security settings` since this comment doesn't actually "document" anything. And then have it tied to the next line without a newline in between

server.js
```diff
@@ -18,8 +20,32 @@ import env from "./config/env.generated.json";
...
+}));
+
+app.use(helmet.ieNoOpen());
+
+app.use(helmet.noSniff());
+
 app.use(express.static(path.resolve(__dirname, `dist`)));
 
+
+/**
```
Pomax: probably want this to be `//....` too

js/security-headers.js
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    styleSrc: [
+      `'self'`,
+      `'unsafe-inline'`,
+      `https://code.cdn.mozilla.net`
+    ],
+    imgSrc: [
+      `'self'`,
+      `data:`,
+      `*`
```
Pomax: hmm, does this make the previous two lines irrelevant... the MDN docs are not super clearn, would you know @cadecairos ?

cadecairos: I believe '*' does cover all possible directives.

mmmavis: @Pomax we need to have `*` there (at least for now) so we can have entry thumbnail show. I can remove `'self'` and `data:` from the list.

Pomax: What if we set this to `http:` and `https:`? That way thumbnails from the web work, but no "mystery content" will work

PR #356, opened 2017-02-10
Upgraded to webpack v2

### LINKED ISSUE #352: Upgrade webpack to v2
https://webpack.js.org/guides/migrating/

Because webpack v1 is deprecated.

mmmavis:
Fixes #352

Changes made based on this migration guide: https://webpack.js.org/guides/migrating/

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,3 +1,5 @@
+var webpack = require(`webpack`);
+
 module.exports = {
   context: `${__dirname}`,
   entry: `./main.jsx`,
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -7,19 +9,25 @@ module.exports = {
...
+            options: {
+              presets: [`es2015`, `react`]
+            }
+          }
+        ]
       }
     ]
-  }
+  },
+  plugins: [
```
Pomax: Please uplift first, then add in more functionality to the config as follow-up. This is good to have but **only** for a production build. Source maps are actually super bloaty so it's far better to know which environment we're in and then either minify (for prod) without source maps, or not minify at all (for dev).

webpack.server.config.js
```diff
@@ -23,19 +24,25 @@ module.exports = {
...
+            options: {
+              presets: [`es2015`, `react`]
+            }
+          }
+        ],
       }
     ]
-  }
+  },
+  plugins: [
```
Pomax: why would the server bundle need minification?

webpack.config.js
mmmavis: kk will do

mmmavis:
PR updated! I removed the `plugins` bits.

Pomax:
looks good to me. Also filed https://github.com/mozilla/network-pulse/issues/359

mmmavis:
Merged.

Thanks for filing that followup ticket! @Pomax

PR #362, opened 2017-02-10
'load more' button CSS improvement

mmmavis:
@alanmoo I thought about doing the suggestion you have on https://github.com/mozilla/network-pulse/pull/348#discussion_r100635697 but that means we are wrapping all project cards in ONE `.row` element... which doesn't sound right to me? If we wanna make use of `.row` class we should wrap projects in groups of 3 (can be tackled as a separate PR if needed).

### DIFF
--- a/scss/main.scss
+++ b/scss/main.scss
@@ -176,10 +176,6 @@ button[type="submit"] {
     margin-left: -$project-card-side-margin;
     margin-right: -$project-card-side-margin;
   }
-
-  .view-more {
-    text-align: center;
-  }
 }
 
 // Components
--- a/scss/main.scss
+++ b/scss/main.scss
@@ -31,38 +31,17 @@ p {
 }
 
 .btn-link {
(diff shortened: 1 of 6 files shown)


mmmavis:
PR updated!

scss/main.scss
```diff
@@ -31,38 +31,17 @@ p {
 }
 
 .btn-link {
-  cursor: pointer;
-  padding: 0;
+  &.inline-link {
```
alanmoo: Doesn't this just make it look like a regular anchor element? Can we just use `<a href="/">` with no classes?

mmmavis: We should only use `<a>` for hyperlinks. I can rename the class name if it's confusing.

alanmoo: While I think this could be a hyperlink, this is fine, let's get it merged in. We've got bigger things to deal with.

PR #366, opened 2017-02-15
fixes link button styling used in <HintMessage>

mmmavis:


mmmavis:
@alanmoo I could, but since we are using `classnames` in other places in the codebase I'm gonna stick with that approach here.

PR #372, opened 2017-02-21
"Add new entry" form

mmmavis:
Relates to #309


components/project-card/project-card.jsx
```diff
@@ -149,7 +149,7 @@ export default React.createClass({
                   : (<div>
                       <Link to={detailViewLink} onClick={this.handleReadMoreClick}>Read more</Link>
                     </div>);
-    let creators = this.props.creators ? `By ${this.props.creators.join(`, `)}` : null;
+    let creators = this.props.creators.length > 0 ? `By ${this.props.creators.join(`, `)}` : null;
```
Pomax: there seem to be a lot of PropTypes missing, so it seems like this component shouldn't be able to rely on there being a `props.creators` at all. Do we want to test for creators, then access length, or do we want to make sure `creators` is part of the known proptypes?

components/project-card/project-card.scss
```diff
@@ -51,6 +51,11 @@
     margin: 0.5rem 0;
   }
 
+  p {
+    font-size: 0.9375rem;
```
Pomax: this is a very specific value... why this precise? =D

js/service.js
```diff
@@ -111,10 +111,49 @@ function callURL(route, params = {}) {
...
+  return new Promise((resolve, reject) => {
+    request.open(`POST`, `${pulseAPI}/entries/`, true);
+
+    request.withCredentials = true;
+    request.onload = (event) => {
+      let result = event.currentTarget;
+
+      if (result.status >= 200 && result.status < 400) {
+        try {
+          resolve(JSON.parse(result.response));
```
Pomax: technically this could also fire when `resolve` throws, so a better solution is:
```
let data;

try {
  data = JSON.parse(result.response);
} catch (error) {
  // this error can only have come from JSON parsing
  reject(error);
}

resolve(data);
```

pages/add/add.jsx
```diff
@@ -31,11 +38,117 @@ export default React.createClass({
...
+    Service.nonce()
+      .then((nonce) => {
+        callback(nonce);
+      })
+      .catch((reason) => {
+        this.setState({
+          authError: true
+        });
+        console.error(reason);
+        callback(false);
```
Pomax: why does this need to happen? if the callback needs to work for success and failure, it might make more sense to ensure the callback is a dual `callback(error, data)` function:
```
.then(nonce => {
  callback(false, nonce);
})
.catch(reason => {
  this.setState({ authError: true });
  callback(new Error("could not retrieve data from /nonce"));
});
```

pages/add/add.jsx
```diff
@@ -31,11 +38,117 @@ export default React.createClass({
...
+      });
+  },
+  postEntry(entryData) {
+    this.getNonce((nonce) => {
+      if (!nonce) { return; }
```
Pomax: in line with the above comment, at this point we would then know that this is a real error, not "oh we expected this, return, nothing wrong here" =)

pages/add/form-fields/detail-info-fields.js
```diff
@@ -0,0 +1,108 @@
+module.exports = {
+  creators: {
+    type: `text`,
+    label: `Who are the creators? This could be staff, contributors, partners…`,
+    placeholder: `Name`,
+    fieldClassname: `form-control`,
+    multiplicity: 1,
+    addLabel: `+ Add another`,
```
Pomax: is the "+" intentional here?

pages/add/form-fields/detail-info-fields.js
```diff
@@ -0,0 +1,108 @@
...
+    fieldClassname: `form-control`,
+    validator: {
+      error: `Maximum 300 characters.`,
+      validate: function(value) {
+        if (!value) {
+          // this is an optional field, it can be left blank
+          return;
+        }
+
+        return value && value.length > 300; // return true if value fails validation
```
Pomax: probably better to do this explicit:
```
let passes = value && value.length > 300;
if (!passes) return new Error("blah!");
```
And we don't need a final return, as the test is "does this function return a truthy value", and a natural function end is equivalent to `return undefined`, so is falsey

pages/add/form-fields/detail-info-fields.js
```diff
@@ -0,0 +1,108 @@
...
+  tags: {
+    type: `text`,
+    label: `Keywords to help with search by program, event, campaign, subject …`,
+    placeholder: `#mozfest  #code  #tool`,
+    fieldClassname: `form-control`
+  },
+  issues: {
+    type: `checkboxGroup`,
+    label: `Check any Key Internet Issues that relate to your project.`,
+    options: [ `Online Privacy & Security`, `Open Innovation`, `Decentralization`, `Web Literacy`, `Digital Inclusion` ],
```
Pomax: I'd like to tie this to the api server's `/issues?format=json` but definitely not now

pages/add/form-fields/detail-info-fields.js
```diff
@@ -0,0 +1,108 @@
...
+    // optional: true
+  },
+  'get_involved': {
+    type: `text`,
+    label: `Looking for support? Describe how people can do that.`,
+    placeholder: `Contribute to the code.`,
+    fieldClassname: `form-control`,
+    validator: {
+      error: `Maximum 300 characters.`,
+      validate: function(value) {
```
Pomax: if we have multiple "max X characters" fields, they use the same validation rules, and we can DRY that out

pages/add/form-fields/detail-info-fields.js
```diff
@@ -0,0 +1,108 @@
...
+    fieldClassname: `form-control`,
+    validator: {
+      error: `Not a valid URL. Remember to include protocol (http:// or https://).`,
+      validate: function(value) {
+        if (!value) {
+          // this is an optional field, it can be left blank
+          return;
+        }
+
+        let r = new RegExp(/^((http||https):\/\/)([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/);
```
Pomax: this makes me nervous... what are we testing for here? I'm almost certain that we can use the URL api to test for what we want to test for here.

pages/add/add.jsx
```diff
@@ -31,11 +38,117 @@ export default React.createClass({
...
+        .catch((reason) => {
+          this.setState({
+            serverError: true
+          });
+          console.error(reason);
+        });
+    });
+  },
   getContentForLoggedInUser() {
+    let authErrorMessage = this.state.authError ? <span className="error">You seem to be logged out at this moment. Try <a href={user.getLoginURL(utility.getCurrentURL())}>logging in</a> again?</span> : null;
```
mmmavis: I'm thinking maybe we can temporarily save form values in localstorage so after user is redirected back from login we can restore the form?

(we can file a separate ticket for this)

pages/add/form-fields/detail-info-fields.js
```diff
@@ -0,0 +1,108 @@
...
+    validator: [
+      {
+        error: `Not a valid image URL. Only JPEG, PNG, GIF, or SVG file is allowed. Remember to include protocol (http:// or https://)`,
+        validate: function(value) {
+          if (!value) {
+            // this is an optional field, it can be left blank
+            return;
+          }
+
+          let r = new RegExp(/^((http||https):\/\/)([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?\.(jpg|jpeg|png|gif|svg)$/);
```
Pomax: same note above

Pomax:
Looks like the multiplicity CSS is off a little in Chrome:
screenshot 167

Pomax:
Additional names don't end up below each other, so the "add more" button moves right to left to right to left the more creators you add. Is that intentional? (it feels like it makes the UI unpredictable)

screenshot 168

Pomax:
Tags are called "keywords", and they're shown as using hashtags... Is that the intention? Do we strip the hashtags when we send to the API? Do we present searches based on tags with `#` injected back in? This seems like a thing we should probably change so that we at least tell people not to use hashtags?

screenshot 169

Pomax:
Looks like we're missing some spacing between the checkboxes and labels for our internet issues

screenshot 170

Pomax:
@mmmavis @kristinashu it feels like the submission process here is "too fast", in that there is no post confirmation? You're taken to the issue you just posted, which I remember we wanted to have happen, but it might be a good idea review this from a usability perspective. It feels like once you submit the entry, the view of your just-posted entry should have some kind of "thanks for submitting, here's the thing you posted, all live and everything!"

components/project-card/project-card.scss
mmmavis: I'm not sure. haha this is what's in the Sketch file (15px).

components/project-card/project-card.jsx
mmmavis: yea I can add `PropTypes` to `Details` but we might still need to rely on `this.props.creators.length > 0` to decide if we should render the "By _____" line or not. (as `creator` can be an empty array)

components/project-card/project-card.scss
Pomax: This feels like an arbitrary pixel difference... @kristinashu does this *need* to be 1px below the baseline browser default?

components/project-card/project-card.jsx
mmmavis: ah sorry ignored what I said. I think I'm just gonna change this to 

```
let creators = this.props.creators && this.props.creators.length > 0 ? `By ${this.props.creators.join(`, `)}` : null;
```

components/project-card/project-card.scss
mmmavis: Yea I wasn't sure if it was intended. The paragraph text in (detail view) project card is 15px and with line height set to 1.333. Whereas default body text is 16px + line height 1.5.

kristinashu: I'm ok with rounding off numbers/simplifying...

components/project-card/project-card.jsx
Pomax: also works =)

pages/add/add.jsx
Pomax: Users should not be able to even see a form before they're logged in though

mmmavis: Yea but in case they log themselves out on another tab before the form gets submitted. That's one scenario that I think /nonce can fail.

Pomax: hm, let's file this as potential edge case and deal with as follow-up

mmmavis:
Thanks for the review and comments, @Pomax ! I'm gonna merge this PR now so people can test it on staging tomorrow. I will create individual ticket for every comment you mentioned here tomorrow. Gonna call it a day now 😴

PR #379, opened 2017-02-22
MoFo unified footer

### LINKED ISSUE #178: Use universal footer
Consider including: 
- Legal
- GitHub Repo
- About Pulse
- Mozilla logo, description, and link
- Contact email?

mmmavis:
Fixes #178

### DIFF
--- a/components/footer/footer.jsx
+++ b/components/footer/footer.jsx
@@ -1,16 +1,47 @@
 import React from 'react';
-// import { Link } from 'react-router';
+import { MofoFooter } from 'mofo-ui';
 
-export default React.createClass({
-  render() {
-    return (
-      <footer>
-        <div className="container">
-          <p>Get new projects by email. <a id="sign-up-btn" href="https://groups.google.com/a/mozillafoundation.org/d/forum/pulse-notifications/join">Sign up</a>.</p>
-          <p><a href="https://github.com/mozilla/network-pulse">Contribute</a> to this prototype.</p>
-          <p>Submit an <a href="https://docs.google.com/a/mozillafoundation.org/forms/d/e/1FAIpQLScsZ6NLDIo87YH8Z3guR-9slPampPTavcKgbqOJvLsrnqIVMQ/viewform">Impact Story tip</a>.</p>
-        </div>
-      </footer>
-    );
-  }
-});
+const Footer = () => {
+  const FOOTER_LINKS = [
+    {
+      iconType: `github`,
+      link: `https://github.com/mozilla/network-pulse`,
+      text: `Github`
+    },
+    {
+      iconType: `chat`,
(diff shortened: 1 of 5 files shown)


mmmavis:
Nice! Dropping in mofo-ui and it works like a charm! @alanmoo 👀 ?

scss/main.scss
```diff
@@ -93,14 +93,11 @@ a {
 }
 
 #main {
-  padding-bottom: 160px;
+  padding-top: 30px;
+  padding-bottom: 420px;
```
alanmoo: Why is this necessary? Looks like `footer` has absolute positioning for some reason. Removing that negates the need for this, is that something that can be done?

components/footer/footer.jsx
```diff
@@ -1,16 +1,47 @@
...
-          <p>Get new projects by email. <a id="sign-up-btn" href="https://groups.google.com/a/mozillafoundation.org/d/forum/pulse-notifications/join">Sign up</a>.</p>
-          <p><a href="https://github.com/mozilla/network-pulse">Contribute</a> to this prototype.</p>
-          <p>Submit an <a href="https://docs.google.com/a/mozillafoundation.org/forms/d/e/1FAIpQLScsZ6NLDIo87YH8Z3guR-9slPampPTavcKgbqOJvLsrnqIVMQ/viewform">Impact Story tip</a>.</p>
-        </div>
-      </footer>
-    );
-  }
-});
+const Footer = () => {
+  const FOOTER_LINKS = [
```
alanmoo: Should we should have a set of default links in MoFo UI so this isn't always necessary?

mmmavis: Um not sure, I think links can vary a lot across different sites.

scss/main.scss
mmmavis: We want the footer to always stay at the page bottom. Without reserving additional space in `#main` will result in something like

screen shot 2017-02-22 at 3 32 07 pm

vs with reserved space

screen shot 2017-02-22 at 3 32 26 pm

I can't think of a better solution without having annoying hardcoded reserved space to achieve this. If we don't have absolute poisoning on `footer`, `footer` can end up being in a middle of the page on large devices. Plus we are loading dynamic content on most of the pages - that adds in some complexity too. Suggestion?

components/footer/footer.jsx
mmmavis: Um now I think of it, it could be helpful if we have defaults for `cc-license`, `code-of-conduct`, `cookies`, `legal`, and `privacy`. Let's move this discussion to mofo-ui: https://github.com/mozilla/mofo-ui/issues/55

scss/main.scss
alanmoo: Ah, right- short pages with little content. Fair enough.

PR #407, opened 2017-03-01
Registered components as user class's listeners

### LINKED ISSUE #343: add "user.verify" to componentWillUnmount
This will allow things like "failedLogin" to reset between different pages.

mmmavis:
Fixes #343 

I pretty much just stole your suggested code @Pomax  😁

### DIFF
--- a/pages/bookmarks.jsx
+++ b/pages/bookmarks.jsx
@@ -21,14 +21,22 @@ export default React.createClass({
   handleLogOutBtnClick(event) {
     event.preventDefault();
     user.logout();
-    this.setState({ user });
   },
   componentDidMount() {
     // get IDs of user's bookmarked entries
     this.setState({bookmarks: getBookmarks()});
 
-    // let's verify what the logged in status is
-    user.verify(this.props.router.location, () => this.setState({ user }));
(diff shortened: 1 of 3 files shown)


pages/bookmarks.jsx
```diff
@@ -21,14 +21,22 @@ export default React.createClass({
...
     user.logout();
-    this.setState({ user });
   },
   componentDidMount() {
     // get IDs of user's bookmarked entries
     this.setState({bookmarks: getBookmarks()});
 
-    // let's verify what the logged in status is
-    user.verify(this.props.router.location, () => this.setState({ user }));
+    user.addListener(this);
```
mmmavis: I wonder if I should import `app-user.js` as `User` (instead of `user`) since "user" is a class?

Pomax: pretty sure it's not, based on https://github.com/mozilla/network-pulse/blob/2ad7cee35201d9c9bad6e98bf217b2a5b3eb697d/js/app-user.js#L166 -- we create a singleton instance of the User class, and everything then gets to use that single instance as a shared object.

Issue #387, opened 2017-02-23
improve load time by restricting page_size for `entries` to 24


Pomax:
Right now page_size is set to 996 which makes the request transfer the entire database of entries on every call. Let's cut this down to 24 instead, so it doesn't take a really long time to do a network fetch.

xmatthewx:
This was discussed before with @alanmoo and @mmmavis. I believe there was a concern about slow loads on Load More. Or repeat searches queries. 

How big approximately is a blobby blob of json per 100 entries? What's a reasonable amount of kb to transfer?

Pomax:
you don't really wan to transfer more than 1kb. I'm going to be pragmatic here and say "just load what you need to show the user what they're supposed to see". If the screen fits 24 items, load 24 items, at mot prefetch the next 24, and keep network activity lean.

xmatthewx:
That sounds reasonable. 

Is there a way I can do a throttled test on my end by viewing /latest?page_size=24 manually? It seems to be fetching all still with that parameter. Site is slow on 3G, but Load More seems instant.

Pomax:
yeah if you want to check the difference, try these two links:

- full: https://network-pulse-api-staging.herokuapp.com/entries/?format=json
- limited: https://network-pulse-api-staging.herokuapp.com/entries/?format=json&page_size=12&page=3

And then you can play with the "page size" value to see the difference

mmmavis:
Fixed in https://github.com/mozilla/network-pulse/pull/410

Issue #361, opened 2017-02-10
Refactor project-list component


alanmoo:
The project-list component seems to be doing a bit too much, causing some issues with the different ways it's used requiring band-aids.

Let's simplify the API of it, here's my thought:

`Project List` gets passed in 2 props: `projects`, which is an array of objects (entries from the API), and `sortBy`, a key to sort the projects by or a list of IDs by which to order them.

Any other logic in there gets pulled up a level (like the `issues` bit) so that those higher components handle fetching the data and passing it in to the `Project List`. This should also allow you to DRY out the `search.jsx` and `project-list.jsx` files.

In addition, the CSS could be cleaned up a bit to use bootstrap classes and less custom styles.

mmmavis:
Fixed in PR https://github.com/mozilla/network-pulse/pull/410.

PR #410, opened 2017-03-02
Project list refactor. Fetch 24 projects at a time by default.

mmmavis:
Related to #361, #387

### DIFF
--- a/components/project-list/project-list.jsx
+++ b/components/project-list/project-list.jsx
@@ -14,28 +14,27 @@ export default React.createClass({
     moreEntriesToFetch: React.PropTypes.bool.isRequired,
     fetchData: React.PropTypes.func.isRequired
   },
-  handleViewMoreClick() {
-    this.props.fetchData();
-  },
   renderProjectCards() {
     return this.props.entries.map(project => {
       return <ProjectCard key={project.id} {...Utility.processEntryData(project)} />;
     });
   },
   renderLoadingNotice() {
+    if (!this.props.loadingData ) return null;
+
     // 3 empty <div></div> here are for the loading animation dots (done in CSS) to show.
-    return this.props.loadingData ?
-      (<div className="loading my-5 d-flex justify-content-center align-items-center">
-        <div></div>
-        <div></div>
-        <div></div>
-      </div>) : null;
+    return <div className="loading my-5 d-flex justify-content-center align-items-center">
+              <div></div>
+              <div></div>
+              <div></div>
+            </div>;
   },
   renderViewMoreBtn() {
-    return this.props.moreEntriesToFetch ?
-      (<div className="view-more text-center">
-        <button type="button" className="btn btn-outline-info" onClick={this.handleViewMoreClick}>View more</button>
-       </div>) : null;
+    if (!this.props.moreEntriesToFetch) return null;
+
+    return <div className="view-more text-center">
+            <button type="button" className="btn btn-outline-info" onClick={this.props.fetchData}>View more</button>
+           </div>;
   },
   render() {
     return (<div className="project-list">
--- a/components/project-list/project-list.jsx
+++ b/components/project-list/project-list.jsx
@@ -1,113 +1,49 @@
 import React from 'react';
 import ProjectCard from '../project-card/project-card.jsx';
-import Service from '../../js/service.js';
 import Utility from '../../js/utility.js';
 
 export default React.createClass({
-  numProjectsInBatch: 24, // make sure this number is divisible by 2 AND 3 so rows display evenly for different screen sizes
-  getInitialState() {
-    return {
-      dataLoaded: false,
-      entries: [],
-      displayBatchIndex: 1
-    };
-  },
   getDefaultProps() {
     return {
-      params: {
-        featured: null,
-        issue: null,
-        ids: null
-      }
+      entries: []
     };
   },
-  componentDidMount() {
-    this.fetchData(this.props.params);
-  },
-  componentWillReceiveProps(nextProps) {
-    // Reset state
-    this.resetState();
-
-    // <ProjectList> doesn't always get unmounted and remounted when navigating between pages.
-    // (e.g., When navigating between the /issue/issue-name pages <Issue><ProjectList></Issue>
-    //        do not get re-mounted since the same components are being rendered.)
-    // It is treated as passing new props to <ProjectList {...newProps} />
-    let newIssueEntered = (nextProps.params.issue !== this.props.params.issue);
-
-    if ( newIssueEntered ) {
-      // Fetch data based on the new params props to ensure data gets fetched and displayed accordingly.
-      this.fetchData(nextProps.params);
-    }
-  },
-  resetState() {
-    this.setState(this.getInitialState());
-  },
-  fetchData(params = {}, entries = [], apiPageIndex = 1) {
-    params.page = apiPageIndex;
-
-    Service.entries
-      .get(params)
-      .then((data) => {
-        entries = entries.concat(data.results);
-
-        if (data.next) { // there are more data in the database we need to fetch
-          apiPageIndex += 1;
-          this.fetchData(params);
-        } else {
-          this.setState({
-            dataLoaded: true,
-            entries: entries
-          });
-        }
-      })
-      .catch((reason) => {
-        console.error(reason);
-      });
-  },
-  applyFilterToList(entries,params = {}) {
-    if ( params.ids ) { // we want to show the list from the most recently bookmarked entry first
-      let bookmarkedIdArray = params.ids.split(`,`);
-
-      return entries.sort((a,b) => {
-        return bookmarkedIdArray.indexOf(a.id.toString()) > bookmarkedIdArray.indexOf(b.id.toString());
-      });
-    } else {
-      return entries;
-    }
-
+  propTypes: {
+    entries: React.PropTypes.array.isRequired,
+    loadingData: React.PropTypes.bool.isRequired,
+    moreEntriesToFetch: React.PropTypes.bool.isRequired,
+    fetchData: React.PropTypes.func.isRequired
   },
   handleViewMoreClick() {
-    this.setState({displayBatchIndex: this.state.displayBatchIndex+1});
+    this.props.fetchData();
+  },
+  renderProjectCards() {
+    return this.props.entries.map(project => {
+      return <ProjectCard key={project.id} {...Utility.processEntryData(project)} />;
+    });
+  },
+  renderLoadingNotice() {
+    // 3 empty <div></div> here are for the loading animation dots (done in CSS) to show.
+    return this.props.loadingData ?
+      (<div className="loading my-5 d-flex justify-content-center align-items-center">
+        <div></div>
+        <div></div>
+        <div></div>
+      </div>) : null;
+  },
+  renderViewMoreBtn() {
+    return this.props.moreEntriesToFetch ?
+      (<div className="view-more text-center">
+        <button type="button" className="btn btn-outline-info" onClick={this.handleViewMoreClick}>View more</button>
+       </div>) : null;
   },
   render() {
-    let projects = (<div className="loading my-5 d-flex justify-content-center align-items-center">
-                      <div></div>
-                      <div></div>
-                      <div></div>
-                    </div>);
-    let moreToShow;
-    let showViewMoreBtn;
-
-    if (this.state.dataLoaded) {
-      // get all matching projects
-      projects = this.applyFilterToList(this.state.entries,this.props.params);
-      // we only want to show a fixed number of projects at once (this.numProjectsInBatch)
-      // first, check to see if there are more projects to show after this batch
-      moreToShow = (projects.length/this.numProjectsInBatch) > this.state.displayBatchIndex;
-      showViewMoreBtn = moreToShow ? (<div className="view-more text-center">
-                                        <button type="button" className="btn btn-outline-info" onClick={this.handleViewMoreClick}>View more</button>
-                                      </div>)
-                                   : null;
-      // prepare ProjectCards we are going to render in this batch
-      projects = projects.slice(0,this.state.displayBatchIndex*this.numProjectsInBatch).map((project) => {
-        return ( <ProjectCard key={project.id} {...Utility.processEntryData(project)} /> );
-      });
-      projects = <div className="projects">{projects}</div>;
-    }
-
     return (<div className="project-list">
-              { projects }
-              { showViewMoreBtn }
+              <div className="projects">
+                { this.renderProjectCards() }
+              </div>
+              { this.renderLoadingNotice() }
+              { this.renderViewMoreBtn() }
             </div>);
   }
 });
--- a/components/project-loader/project-loader.jsx
+++ b/components/project-loader/project-loader.jsx
@@ -10,8 +10,7 @@ export default React.createClass({
       entries: [],
       moreEntriesToFetch: false,
       totalMatched: 0,
-      params: {},
-      existingPromise: null
+      tokenOfExistingPromise: null
     };
   },
   getDefaultProps() {
@@ -24,10 +23,10 @@ export default React.createClass({
   },
   componentWillReceiveProps(nextProps) {
     // Reset state before fetching data for the new params.
-    // We want keep existingPromise on record for fetchData
+    // We want to keep existingPromise on record for fetchData
     // to handle it accoringly.
-    let existingPromise = this.state.existingPromise;
-    let newState = Object.assign(this.getInitialState(), {existingPromise: existingPromise});
+    let tokenOfExistingPromise = this.state.tokenOfExistingPromise;
+    let newState = Object.assign(this.getInitialState(), {tokenOfExistingPromise: tokenOfExistingPromise});
 
     this.setState(newState, () => {
       this.fetchData(nextProps.params);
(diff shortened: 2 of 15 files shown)


components/project-list/project-list.jsx
```diff
@@ -1,113 +1,49 @@
 import React from 'react';
```
mmmavis: I refactored this into 2 components:

- The new `<ProjectList>` component: (`components/project-list/project-list.jsx`)
  - Render things based on what's been passed to it.

- `<ProjectLoader>` component: (`components/project-loader/project-loader.jsx`)
  - Fetches data and pass it to `<ProjectList`> for rendering UI.

js/service.js
```diff
@@ -1,11 +1,11 @@
 import env from "../config/env.generated.json";
+import BluebirdPromise from "bluebird";
```
mmmavis: We can probably just import `bluebird` as `Promise`. Since other functions in this file use the build-in `Promise` and I don't want to introduce more changes to this already-huge PR, I'm importing `bluebird` as `BluebirdPromise` for now. We can file follow-up PR to clean up this file.

js/service.js
```diff
@@ -90,12 +100,11 @@ function getDataFromURL(route, params = {}) {
  * @param  {Object} params A key-value object to be serialized as a query string
  * @returns {Promise} A promise to resolve an XHR request
  */
-function callURL(route, params = {}) {
```
mmmavis: We don't have `params` here so I removed it.

pages/bookmarks.jsx
```diff
@@ -1,16 +1,23 @@
 import React from 'react';
```
mmmavis: This component is similar to `<ProjectLoader>`, except there are logics to make sure projects we are passing to `<ProjectList>` have already been sorted based on the time they were bookmarked(/stored in `localStorage`).

mmmavis:
@Pomax Sorry for the huge PR. I tried my best to minimize the changes I had to introduce... 🙄 

I added some inline comments... hopefully they'll make reviewing this PR slightly less painful 😆

components/project-list/project-list.jsx
```diff
@@ -1,113 +1,49 @@
...
+    loadingData: React.PropTypes.bool.isRequired,
+    moreEntriesToFetch: React.PropTypes.bool.isRequired,
+    fetchData: React.PropTypes.func.isRequired
   },
   handleViewMoreClick() {
-    this.setState({displayBatchIndex: this.state.displayBatchIndex+1});
+    this.props.fetchData();
+  },
+  renderProjectCards() {
+    return this.props.entries.map((project) => {
```
Pomax: small nit: there is no need for `()` around the argument, nor around the return

package.json
```diff
@@ -62,6 +65,7 @@
     "babel-loader": "^6.2.9",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.16.0",
+    "bluebird": "^3.4.7",
```
Pomax: do we actually need bluebird at this point? I thought `Promise` was supported by every modern browser now (http://caniuse.com/#search=promise)

js/service.js
Pomax: Do we even need this? Except for IE11, everything has Promise built in now.

pages/bookmarks.jsx
```diff
@@ -25,11 +32,44 @@ export default React.createClass({
...
+          moreEntriesToFetch: (this.state.batchIndex+1)*PROJECT_BATCH_SIZE < this.state.bookmarkedIds.length
+        });
+      })
+      .catch((reason) => {
+        console.error(reason);
+      });
+  },
+  sortEntriesByTimeBookmarked(entries) {
+    return entries.sort((a,b) => {
+      return this.state.bookmarkedIds.indexOf(a.id) > this.state.bookmarkedIds.indexOf(b.id);
```
Pomax: probably worth lifting out:
```
let ids = this.state.bookmarkedIds;
let sorter = (a,b) => ids.indexOf(a.id) - ids.indexOf(b.id);
return entries.sort(sorter);
```
Also note that small rewrite: a sorting function returns three numerical values, not a boolean (nevative number for less than, 0 for equal, positive number for greater than)

pages/latest.jsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
-import ProjectList from '../components/project-list/project-list.jsx';
+import ProjectLoader from '../components/project-loader/project-loader.jsx';
 
 export default React.createClass({
```
Pomax: do we even need a full component here or can we just make this a stateless function?
```
export default function (props) {
  return <div><ProjectLoader/></div>;
};
```

pages/issue.jsx
```diff
@@ -1,12 +1,10 @@
...
 import IssueSelector from '../components/issue-selector/issue-selector.jsx';
-import ProjectList from '../components/project-list/project-list.jsx';
+import ProjectLoader from '../components/project-loader/project-loader.jsx';
 
 export default React.createClass({
-  getIssueName() {
-    let issue = ``;
-
-    switch(this.props.params.issue) { // the five issues we have
+  encodeIssueAsUri(issue) {
```
Pomax: does this need to be a full component or can we make this stateless?

pages/search/search.jsx
```diff
@@ -34,7 +29,7 @@ export default React.createClass({
       // make sure input box value is updated with searchQueryInUrl
       // then use searchQueryInUrl as the search param to fetch data from Pulse API
       this.refs.searchInput.setState({value: searchQueryInUrl ? decodeURIComponent(searchQueryInUrl) : ``});
```
Pomax: This feels wrong. State is private, we should not be messing with another component's state. If we want their values to change, in the `render()` function make sure to pass the new value as a prop, and make that component do the right thing when it sees a property update.

mmmavis: Totally agree!

js/service.js
mmmavis: I was looking for a way to abort a Promise so that on the search page previous search result won't accidentally interfere later searches. `bluebird` has this feature handy so I used it. See `fetchData()` method in `project-loader.jsx`.

mmmavis: let me try and see if i can write something myself instead of using `bluebird` just for this.

Pomax:
Hm, I seem to get some console errors on the "favs" tab:
```
Internal Server Error: /entries/
Traceback (most recent call last):
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\django\core\handlers\exception.py", line 39, in inner
    response = get_response(request)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\django\core\handlers\base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\django\core\handlers\base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\django\views\decorators\csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\django\views\generic\base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\rest_framework\views.py", line 477, in dispatch
    response = self.handle_exception(exc)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\rest_framework\views.py", line 437, in handle_exception
    self.raise_uncaught_exception(exc)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\rest_framework\views.py", line 474, in dispatch
    response = handler(request, *args, **kwargs)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\rest_framework\generics.py", line 240, in get
    return self.list(request, *args, **kwargs)
  File "c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\venv\lib\site-packages\rest_framework\mixins.py", line 40, in list
    queryset = self.filter_queryset(self.get_queryset())
  File "C:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\pulseapi\entries\views.py", line 182, in get_queryset
    ids = [ int(x) for x in ids.split(',') ]
  File "C:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api\pulseapi\entries\views.py", line 182, in <listcomp>
    ids = [ int(x) for x in ids.split(',') ]
ValueError: invalid literal for int() with base 10: ''
[13/Mar/2017 16:16:32] "GET /userstatus/?ordering=-created&page_size=24&format=json HTTP/1.1" 200 75
```
Which seems to come from the client asking for:
```
GET http://test.example.com:8000/entries/?ids=&page=1&ordering=-created&page_size=24&format=json 500 (Internal Server Error)
```
Where is that `ids=` without a value coming from? That query argument shouldn't be there if we're not requesting specific entries..

components/project-list/project-list.jsx
```diff
@@ -1,113 +1,49 @@
...
+      (<div className="loading my-5 d-flex justify-content-center align-items-center">
+        <div></div>
+        <div></div>
+        <div></div>
+      </div>) : null;
+  },
+  renderViewMoreBtn() {
+    return this.props.moreEntriesToFetch ?
+      (<div className="view-more text-center">
+        <button type="button" className="btn btn-outline-info" onClick={this.handleViewMoreClick}>View more</button>
```
Pomax: Is there a reason we don't say `onClick={this.props.fetchData}`?

components/project-list/project-list.jsx
```diff
@@ -1,113 +1,49 @@
...
+    // 3 empty <div></div> here are for the loading animation dots (done in CSS) to show.
+    return this.props.loadingData ?
+      (<div className="loading my-5 d-flex justify-content-center align-items-center">
+        <div></div>
+        <div></div>
+        <div></div>
+      </div>) : null;
+  },
+  renderViewMoreBtn() {
+    return this.props.moreEntriesToFetch ?
```
Pomax: I'd return early, rather than use a tertiary here:
```
if (!this.props.moreEntriesToFetch) return null;
return (
  <div>
    ...
  </div>
);
```

components/project-list/project-list.jsx
```diff
@@ -1,113 +1,49 @@
...
-    this.setState({displayBatchIndex: this.state.displayBatchIndex+1});
+    this.props.fetchData();
+  },
+  renderProjectCards() {
+    return this.props.entries.map(project => {
+      return <ProjectCard key={project.id} {...Utility.processEntryData(project)} />;
+    });
+  },
+  renderLoadingNotice() {
+    // 3 empty <div></div> here are for the loading animation dots (done in CSS) to show.
```
Pomax: I'd return early, rather than use a tertiary here:
```
if (!this.props.loadingData ) return null;
return (
  <div>
    ...
  </div>
);
```

components/project-loader/project-loader.jsx
```diff
@@ -0,0 +1,87 @@
...
+  },
+  getDefaultProps() {
+    return {
+      params: {}
+    };
+  },
+  componentDidMount() {
+    this.fetchData();
+  },
+  componentWillReceiveProps(nextProps) {
```
Pomax: I don't quite understand what you're doing in this function. Is there a reason we can't track this promise as a perfectly normal `this.promiseToken`? Why do we need to put it in state?

components/project-loader/project-loader.jsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    let combinedParams = Object.assign({}, params, { page: this.state.apiPageIndex });
+    let token = {};
+
+    this.setState({
+      // we are about to make a new request, set loadingData to true.
+      loadingData: true,
+      // we want component to know about the promise returned from the to-be-called Service.entries.get()
+      tokenOfExistingPromise: token
```
Pomax: Why are we putting this here, instead of on `this`? We don't use it for any UI rendering purposes, there is no need to have it live in state, especially as it needs to be modified outside of the React lifecycle functions.

components/project-loader/project-loader.jsx
```diff
@@ -0,0 +1,87 @@
...
+      loadingData: false,
+      entries: this.state.entries.concat(data.results),
+      apiPageIndex: data.next ? this.state.apiPageIndex+1 : this.state.apiPageIndex,
+      moreEntriesToFetch: !!data.next,
+      totalMatched: data.count
+    });
+  },
+  renderSearchResult() {
+    return (this.props.params.search && !this.state.loadingData ) ?
+      (<p>{this.state.totalMatched} result{this.state.totalMatched > 1 ? `s` : ``} found for ‘{this.props.params.search}’</p>) : null;
```
Pomax: this feels a little too tertiary...
```
if (!this.props.params.search || this.state.loadingData) return null;

var total = this.state.totalMatched,
    plural = (total === 0 || total > 1), // because "0 results"
    term = this.props.params.search,
    searchResultNotice = `${total} result${plural ? 's' : ''} found for ‘${term}’`;

return <p>{searchResultNotice}</p>;
```

js/build-env-config.js
```diff
@@ -15,7 +15,13 @@ if (environment !== `PRODUCTION`) {
 const HOST = habitat.get(`HOST`);
 const PORT = habitat.get(`PORT`);
 const ORIGIN = HOST + (PORT ? `:${PORT}` : ``);
-const CONFIG = { HOST, PORT, ORIGIN, PULSE_API: habitat.get(`PULSE_API`)};
+const CONFIG = {
+  HOST,
+  PORT,
+  ORIGIN,
+  PROJECT_BATCH_SIZE: habitat.get(`PROJECT_BATCH_SIZE`) || 24,
```
Pomax: `get` takes a fallback value as 2nd arg: ```habitat.get(`PROJECT_BATCH_SIZE`, 24)```

pages/bookmarks.jsx
Pomax: does that not mean we can just re-use `<ProjectLoader>` but with an optional attribute `sortOnLoad=....` so that we don't need this separate component, but we can simply make ProjectLoader either force-sort if there is a `this.props.sortOnLoad`, or do what it normally does if there isn't?

pages/featured.jsx
```diff
@@ -1,10 +1,7 @@
 import React from 'react';
-import ProjectList from '../components/project-list/project-list.jsx';
+import ProjectLoader from '../components/project-loader/project-loader.jsx';
+
+export default function () {
```
Pomax: Do we want to future-proof this? 
```
export default function(props) {
  return <div><ProjectLoader {...props} params={{featured: `True`}} /></div>;
}
```

pages/featured.jsx
```diff
@@ -1,10 +1,7 @@
...
+  return <div><ProjectLoader params={{featured: `True`}} /></div>;
```
Pomax: why is there a `<div>` around this? Shouldn't the ProjectLoader return its own content wrapper with associated className?

pages/latest.jsx
```diff
@@ -1,10 +1,6 @@
...
+import ProjectLoader from '../components/project-loader/project-loader.jsx';
 
-export default React.createClass({
-  render() {
-    return (
-      <ProjectList />
-    );
-  }
-});
+export default function () {
```
Pomax: same future proof question. Also, clearly latest.jsx is identical to featured.jsx except for the featured flag, so I'm not sure there's much point in creating separate components for them - why don't we just use `<ProjectLoader/>` and `<ProjectLoader featured={true}/>` where we now import these?

pages/featured.jsx
Pomax: why are we not bootstrapping as `<ProjectLoader featured={true} />`? Passing a string into an object that is bound to "params" feels way too many levels of indirection.

pages/search/search.jsx
```diff
@@ -3,148 +3,67 @@ import { browserHistory } from 'react-router';
...
   componentWillReceiveProps(nextProps) {
     // when window.history.back() or windows.history.forward() is triggered
     // (e.g., clicking on browser's back / forward button)
     // we want to make sure search result gets updated accordingly
-    let searchKeyword = nextProps.location.query.keyword;
-    this.searchByQueryInUrl(searchKeyword);
+    this.setState({ keywordSearched: nextProps.location.query.keyword });
   },
   componentDidMount() {
     let queryKeyword = this.props.location.query.keyword;
```
Pomax: Given that all the other code has been removed, why the var capture?
```
this.setState({
  keywordSearched: this.props.location.query.keyword
});
```

pages/bookmarks.jsx
mmmavis: hey @Pomax can I suggest that we open a followup ticket for this? I would like to have this already-huge PR merged first if possible. I'll get to the followup ticket right after this PR is merged.

components/project-loader/project-loader.jsx
mmmavis: This has been fixed. Not sure why GitHub is still showing this comment 😁

components/project-loader/project-loader.jsx
```diff
@@ -3,14 +3,14 @@ import Service from '../../js/service.js';
 import ProjectList from '../project-list/project-list.jsx';
 
 export default React.createClass({
+  promiseToken: null,
```
Pomax: this is now a prototype-leve property: better to not put it here, and instead explicitly rely on `this.promiseToken` in the code (exploiting the fact that if there none declared yet, `this.promiseToken` is `undefined`, which is falsey)

components/project-loader/project-loader.jsx
```diff
@@ -37,19 +35,17 @@ export default React.createClass({
...
     let token = {};
 
-    this.setState({
-      // we are about to make a new request, set loadingData to true.
-      loadingData: true,
-      // we want component to know about the promise returned from the to-be-called Service.entries.get()
-      tokenOfExistingPromise: token
-    });
+    // we want component to know about the promise returned from the to-be-called Service.entries.get()
+    this.promiseToken = token;
```
Pomax: this is still a little strange: if we're binding this early, then we don't even need `token`, we can just set `this.promiseToken = {}` and then on line 51 say `.get(combinedParams, this.promiseToken)`. However, if that token needs to be "fresh" until the Service chain resolves, we don't actually want to say `this.promiseToken = token` here, yet.

mmmavis:
🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻

pages/bookmarks.jsx
mmmavis: ticket filed: https://github.com/mozilla/network-pulse/issues/429

PR #411, opened 2017-03-02
Added 'deployment' section to README

### LINKED ISSUE #408: Update README.md with deployment flow
- staging gets auto updated after something gets merged into `master` branch

- prod push is manual"deploy master to production"
screen shot 2017-03-02 at 10 20 51 am

mmmavis:
Fixes #408


Pomax:
Concretely, we need to do a deployment rather than a "promotion" because this is a client-side bundle that gets built during deploy, and relies on environment variables at build time. As the staging environment and production environment differ in environment variables, we can't build on staging and then promote to production, as that would simply make our staging client "live"

alanmoo:
Yeah, paste that into the Readme.

Issue #365, opened 2017-02-15
Change "Sign In" <button> on /favs page to <a>


mmmavis:
This is more for code improvement.

https://github.com/mozilla/network-pulse/blob/master/pages/bookmarks.jsx#L39-L41. The "sign in" button on /favs page should probably be an anchor instead. Since ultimately it's a link that redirects to the external login URL (see https://github.com/mozilla/network-pulse-api#get-loginoriginal_urlurl).

Pomax:
Technically it's not a navigation link: users are not intentionally going to "a url", they are intentionally logging in, which happens to use redirects but they couldn't care less really, so in this case a button is fine, as long as it can be tabbed to for accessibility purposes, so I think what we want here is to preserve the button, but make sure it has a `tabIndex=0` so that it's a keyboard-navigable element in the browser.

mmmavis:
Based on [our discussion](https://github.com/mozilla/network-pulse/pull/415#discussion_r105238840) current code is good as is. Gonna close this ticket now.

PR #415, opened 2017-03-05
Changed "Sign In" <button> on /favs page to <a>

Sudip7:
my first pull request to this project. Thanks for  all your help and time.

### DIFF
--- a/pages/bookmarks.jsx
+++ b/pages/bookmarks.jsx
@@ -37,7 +37,7 @@ export default React.createClass({
     return (<p>Only Mozilla staff can login now as we test this new platform. Meanwhile, your favorited projects can still be remembered on this device.</p>);
   },
   getAnonymousContent() {
-    return (<p>Are you Mozilla Staff? <button className="btn btn-link inline-link" onClick={this.handleSignInBtnClick}>Sign in</button>.</p>);
+    return (<p>Are you Mozilla Staff? <a href="javascript:void(0)" onClick={this.handleSignInBtnClick}>Sign in</a>.</p>);
   },
   getloggedInStatusContent() {
     if (user.loggedin) {


pages/bookmarks.jsx
```diff
@@ -37,7 +37,7 @@ export default React.createClass({
     return (<p>Only Mozilla staff can login now as we test this new platform. Meanwhile, your favorited projects can still be remembered on this device.</p>);
   },
   getAnonymousContent() {
-    return (<p>Are you Mozilla Staff? <button className="btn btn-link inline-link" onClick={this.handleSignInBtnClick}>Sign in</button>.</p>);
+    return (<p>Are you Mozilla Staff? <a href="javascript:void(0)" onClick={this.handleSignInBtnClick}>Sign in</a>.</p>);
```
Pomax: This is not a good solution to the problem of a button vs. a real link - `href="javascript:void(0)"` has all kinds of problems; CSP will disallow `javascript:` for instance, and an `<a>` in HTML5 **has** to point somewhere. It can't be used as a placeholder for a thing that pretends to be a link anymore: if it's an `<a>` it *has* to point to a URI (either a `#fragmentid` or an external URL. Empty href, or `javascript:` or just `#` are not correct use of the element anymore)

So instead the solution is to make this a *true* link, so that the link looks like:
```
<a href={ this.createLoginLink }>Sign in</a>
```
so no onClick but instead a function that generates the URL that users need to be redirected too, currently generated somewhere inside of `handleSignInBtnClick` (see https://github.com/Sudip7/network-pulse/blob/4fbbb73ff6956e327a930ebc5dcd11e2ab6899b5/pages/bookmarks.jsx#L19 where `user.login` happens. Instead of that function doing the redirect, that functionality needs to now become a URL that the `<a>` handles instead.

Sudip7: Thanks @Pomax. I don't have much ideas on UI part. But will try learn them as I go. I will update the code n resubmit.

Pomax: perhaps @mmmavis can also weigh in - I think part of the question here is also "do we *need* an `<a>` or is a button actually fine"

mmmavis: @Pomax this is the ticket this PR tries to fix: https://github.com/mozilla/network-pulse/issues/365 When I created that ticket I was thinking it *should* be an `<a>` since it ultimately redirects users to an external link.

Pomax: semantics becomes muddled here: it's not a *navigation* link, users are not intentionally going to "a url", they are intentionally logging in, which happens to use redirects but they couldn't care less really, so in this case a button is fine, as long as it can be tabbed to for accessibility purposes, so I think what we want here is to preserve the button, but make sure it has a `tabIndex=0` so that it's a keyboard-navigable element in the browser.

mmmavis: @Pomax `<button>` should already be tabbable and can be triggered by `space` key. Verified on https://www.mozillapulse.org/favs . Do we still need `tabIndex` if that's the case?

Pomax: Fair point: I would say no. This might not be an issue anymore

mmmavis: Hey sorry @Sudip7 . Based on our discussion above it turns out the current code is good as is - we don't have to change it to `<a>`. As a result I'm gonna close this PR as well as the original ticket that I filed https://github.com/mozilla/network-pulse/issues/365 . I'll try to find an intro ticket and make sure it *is valid* before asking someone to tackle it. Sorry for the confusion. 😁

mmmavis:
Based on [our discussion](https://github.com/mozilla/network-pulse/pull/415#discussion_r105238840) current code is good as is. Gonna close this PR now.

Issue #426, opened 2017-03-14
Clicking on dismiss button should bring focus back to input box


mmmavis:
originally mentioned in https://github.com/mozilla/network-pulse/issues/419#issue-212218068 by @xmatthewx 

> Action: Tap X to clear search. 
Result: Button keeps active state. 
Expected: Focus should go back into input

mmmavis:
Fixed. https://www.mozillapulse.org/search

PR #431, opened 2017-03-15
Input box focus on search page

### LINKED ISSUE #425: Search page: focus should be set to input box by default
originally mentioned in https://github.com/mozilla/network-pulse/issues/419#issue-212218068 by @xmatthewx 

> Action: Nav to search view. 
Expected: Focus should be set to input

mmmavis:
Fixes #425, #426

### DIFF
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -19,6 +19,9 @@ export default React.createClass({
   },
   componentDidMount() {
     this.setState({ keywordSearched: this.props.location.query.keyword });
+    // to select the <input /> rendered from <DebounceInput /> we have to select it by its id.
+    // this.refs.name_for_debounce_input would reference <DebounceInput /> itself, not <input />
+    document.querySelector(`#search-box`).focus();
   },
   updateBrowserHistory() {
     let keywordSearched = this.state.keywordSearched;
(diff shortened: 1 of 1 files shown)


Pomax:
this is a hack. Stick a `ref` on that input and just call `this.refs.search.focus()` -- we should not care in the slightest about whether there is a DOM or not.

pages/search/search.jsx
```diff
@@ -19,6 +19,9 @@ export default React.createClass({
   },
   componentDidMount() {
     this.setState({ keywordSearched: this.props.location.query.keyword });
+    // to select the <input /> rendered from <DebounceInput /> we have to select it by its id.
```
Pomax: no we don't. We can put a `ref="search"` on the `<DebounceInput>` and then we can call whatever function it needs called to focus directly by using `this.ref.search.focus()` (assuming that's the function it exposes)

Pomax:
Having said that, there is no direct way to call this so I've filed https://github.com/nkbt/react-debounce-input/issues/65 in the hopes that there is a way to already do this, but if not hopefully it can be added.

alanmoo:
So in the meantime, R+?

mmmavis:
@Pomax @alanmoo oh yea I agree this is a hack. As Pomax mentioned `focus()` isn't exposed that's why I had to do it this way. I should've filed a ticket on the package's repo though (thanks for doing that @Pomax)

Pomax:
yeah R+ in the mean time, this should not get blocked by a package not exposing what we need

mmmavis:
I updated my inline comments to mention <input>'s `focus()` not being exposed. Also included a link to the ticket Pomax filed.

Gonna merge this PR now.

PR #435, opened 2017-03-20
Image upload

### LINKED ISSUE #424: Add image upload to Input form
This should be tip top priority, before other QA. 

@Pomax has a PR open to enable image upload in the API. https://github.com/mozilla/network-pulse-api/issues/99

We need to hook this up to the existing input form. We will drop the current URL field from the form.  

I'm unsure if this is work for @mmmavis @Pomax or @alanmoo.

mmmavis:
Fixes #424

### DIFF
--- a/pages/add/form/validator.js
+++ b/pages/add/form/validator.js
@@ -69,7 +69,7 @@ const Validator = {
         // there's no file size limit on the backend
         // but it's still good that we enforce a size limit (2MB) on client side
         let base64String = value.base64;
-        let sizeLimit = Math.pow(2, 20) * 2; // 2MB in bytes
+        let sizeLimit = 2097152; // 2MB
         if (base64String && base64String.length > 4/3 * sizeLimit) {
           return new Error(`File size is over 2MB.`);
         }
--- a/pages/add/form/validator.js
+++ b/pages/add/form/validator.js
@@ -36,7 +36,7 @@ const Validator = {
   },
   imageTypeValidator() {
     return {
-      error: `Only JPG, JPEG, PNG, GIF, or SVG file is allowed.`,
+      error: `Only JPG, JPEG, PNG, GIF, or SVG files are accepted.`,
       validate: function(value) {
         if (!value) {
           return;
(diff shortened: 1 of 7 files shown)


pages/add/form/validator.js
```diff
@@ -36,11 +36,16 @@ const Validator = {
   },
   imageTypeValidator() {
     return {
-      error: `Only JPEG, PNG, GIF, or SVG file is allowed.`,
+      error: `Only JPG, JPEG, PNG, GIF, or SVG file is allowed.`,
```
Pomax: "file is" -> "files are", and rather than "allowed" we probably want to say "accepted".

pages/add/form/validator.js
```diff
@@ -36,11 +36,16 @@ const Validator = {
...
-        if (value) {
-          value = value.split(`.`);
-          let fileExtention = value[value.length-1].toLowerCase();
+        if (!value) {
+          return;
+        }
+
+        let fileName = value.name;
+        if (fileName) {
+          fileName = fileName.split(`.`);
```
Pomax: nice trick in modern JS:
```
extension = fileName.substring(fileName.lastIndexOf(".")+1);
```
No need for splitting that way

pages/add/form/validator.js
```diff
@@ -53,6 +58,39 @@ const Validator = {
...
+      error: `File size is over 2MB.`,
+      validate: function(value) {
+        if (!value) {
+          return;
+        }
+
+        // there's no file size limit on the backend
+        // but it's still good that we enforce a size limit (2MB) on client side
+        let base64String = value.base64;
+        let sizeLimit = Math.pow(2, 20) * 2; // 2MB in bytes
```
Pomax: I'd kind of say "let's just use `2097152` because this is literally a fixed number, no need for computation"

mmmavis: Yea I thought about it but couldn't decide. For many devs (like myself) `2097152` isn't a number that we can easily relate to 2MB. I used `Math.pow(2, 20) * 2` so that if we ever need to adjust the size limit in the future, we can simply change a number there and let it compute the final number.

Pomax: just say `// 2mb` as comment. Should be pretty clear?

mmmavis:
🎉

Issue #429, opened 2017-03-14
Code cleanup for bookmarks.jsx


mmmavis:
for context see https://github.com/mozilla/network-pulse/pull/410/files#r105798208

PR #445, opened 2017-03-23
Restore list scroll position

### LINKED ISSUE #428: Remember list scroll position
originally mentioned in https://github.com/mozilla/network-pulse/issues/419#issue-212218068 by @xmatthewx 

> Action: Go back from an entry to a list view. 
Result: list reloads. Expected: show me where I was before

expected behaviour as described in https://github.com/mozilla/network-pulse/pull/445#issuecomment-291942575

> We only need and want to remember scroll position for one specific scenario, when a user goes from List to Detail and back to List. It would likely feel weird in other scenarios. 
> * `Remember position` – Scenario: User opens a Detail view from somewhere in List view, then User triggers browser back from Detail to List view. >>List view displays with prior scroll position. 
 > * `Everything else` – Scenario: User taps Tab (regardless of current and prior position) >> Tab loads from the top

mmmavis:
Fixes #428

### DIFF
--- a/components/project-loader/project-loader.jsx
+++ b/components/project-loader/project-loader.jsx
@@ -1,12 +1,13 @@
 import React from 'react';
 import Service from '../../js/service.js';
 import ProjectList from '../project-list/project-list.jsx';
+import pageSettings from '../../js/app-page-settings';
 
 export default React.createClass({
   getInitialState() {
     return {
       loadingData: false,
-      apiPageIndex: 1,
+      nextApiPageIndex: 1,
       entries: [],
       moreEntriesToFetch: false,
       totalMatched: 0
@@ -18,7 +19,12 @@ export default React.createClass({
     };
   },
   componentDidMount() {
-    this.fetchData();
+    if (pageSettings.restore) {
+      // restore state back to what is stored in pageSettings
+      this.setState(pageSettings.currentList);
+    } else {
+      this.fetchData();
+    }
   },
   componentWillReceiveProps(nextProps) {
     // Reset state before fetching data for the new params.
@@ -30,7 +36,7 @@ export default React.createClass({
     });
   },
   fetchData(params = this.props) {
-    let combinedParams = Object.assign({}, params, { page: this.state.apiPageIndex });
+    let combinedParams = Object.assign({}, params, { page: this.state.nextApiPageIndex });
 
     if (this.promiseToken) {
       // Since we are making asynchronous calls to fetch data,
@@ -56,13 +62,19 @@ export default React.createClass({
       });
   },
   updateStateWithNewData(data) {
-    this.setState({
-      loadingData: false,
+    let currentListInfo = {
       entries: this.state.entries.concat(data.results),
-      apiPageIndex: data.next ? this.state.apiPageIndex+1 : this.state.apiPageIndex,
+      nextApiPageIndex: data.next ? this.state.nextApiPageIndex+1 : this.state.nextApiPageIndex,
       moreEntriesToFetch: !!data.next,
       totalMatched: data.count
-    });
+    };
+
+    // store current project list's info in pageSettings
+    pageSettings.setCurrentList(currentListInfo);
+
+    // update component's state
+    currentListInfo.loadingData = false;
+    this.setState(currentListInfo);
   },
   renderSearchResult() {
     if (!this.props.search || this.state.loadingData) return null;
--- a/components/project-loader/project-loader.jsx
+++ b/components/project-loader/project-loader.jsx
@@ -93,7 +93,8 @@ export default React.createClass({
         <ProjectList entries={this.state.entries}
                     loadingData={this.state.loadingData}
                     moreEntriesToFetch={this.state.moreEntriesToFetch}
-                    fetchData={this.fetchData} />
+                    fetchData={this.fetchData}
+                    restoreScrollPosition={pageSettings.shouldRestore} />
       </div>
     );
   }
--- a/components/project-loader/project-loader.jsx
+++ b/components/project-loader/project-loader.jsx
@@ -19,7 +19,7 @@ export default React.createClass({
     };
   },
   componentDidMount() {
-    if (pageSettings.restore) {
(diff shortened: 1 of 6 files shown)


pages/bookmarks.jsx
```diff
@@ -5,6 +5,7 @@ import ProjectList from '../components/project-list/project-list.jsx';
 import HintMessage from '../components/hint-message/hint-message.jsx';
```
mmmavis: refactor/clean up work for this file will be done as a separate PR: https://github.com/mozilla/network-pulse/issues/429

components/project-loader/project-loader.jsx
```diff
@@ -18,7 +19,12 @@ export default React.createClass({
     };
   },
   componentDidMount() {
-    this.fetchData();
+    if (pageSettings.restore) {
```
Pomax: we should probably call this `shouldRestore` rather than `restore`, as the latter is a verb and so suggests this is a function that should be called.

js/app-page-settings.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+class PageSettings {
+  constructor() {
+    this.reset();
+  }
+
+  reset() {
+    this.currentList = {};
+    this.currentScrollPosition = 0;
+    this.restore = false;
```
Pomax: `shouldRestore` to prevent proptype confusion

js/app-page-settings.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  setCurrentList(list = {}) {
+    this.currentList = list;
+  }
+
+  setScrollPosition() {
+    if (typeof window !== `undefined`) {
+      this.currentScrollPosition = window.pageYOffset;
```
Pomax: maybe add a note here that points out that this will only work as long as entries are not put in their own, scrollable div or the like?

js/app-page-settings.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+
+  setRestore() {
```
Pomax: Not sure we need a function here - calling code can also call `pageSettings.shouldRestore = true` to the same effect.

pages/bookmarks.jsx
```diff
@@ -32,7 +33,12 @@ export default React.createClass({
   componentDidMount() {
     // get IDs of user's bookmarked entries
     this.setState({bookmarkedIds: getBookmarks()}, () => {
-      this.fetchData();
+      if (pageSettings.restore) {
+        // restore state back to what is stored in pageSettings
+        this.setState(Object.assign({},this.state,pageSettings.currentList));
```
Pomax: I don't quite understand this line. `setState` already copies over values from the "newstate" argument into whatever "this.state" is, so the `Object.assign` call here does the same as what `setState` already does. Was there a problem you ran into that necessitated this code?

routes.jsx
```diff
@@ -91,7 +92,7 @@ module.exports = (
       <IndexRoute component={Issues} />
       <Route path=":issue" component={Issue} />
     </Route>
-    <Route path="entry/:entryId" component={Entry} />
+    <Route path="entry/:entryId" component={Entry} onEnter={()=>{ pageSettings.setScrollPosition(); }} onLeave={()=>{ pageSettings.setRestore(); }} />
```
Pomax: `onEnter={evt =>pageSettings.setScrollPosition()}` and `onLeave={evt => pageSettings.setRestore() } />` -- as long as a full arrow function is on one line, `(...) => ...` is equivalent to `function(...) { return ... }`, so as long as the statement after the fat arrow is a single call, you don't need to wrap it in `{}` and you don't need a semi-colon at the end.



(you can use `()` instead of `evt`, but we *know* these functions come with an event handler supplied)

components/project-loader/project-loader.jsx
mmmavis: nice suggestion. i struggled naming that function.

js/app-page-settings.js
mmmavis: thought it would be better to have a function that does it (encapsulation?) though im not 100% certain

pages/bookmarks.jsx
mmmavis: Nice catch! I think originally I had a 3rd argument that later got removed. I forgot to update the rest of the code in that line that's why.

js/app-page-settings.js
Pomax: it's a future-proof in case we ever need to do more than just setting a property to true. But in that case this should really be
```
setRestore(value) {
  value = typeof value === "undefined"? false : value;
  this.shouldRestore = value;
}
```
since a `set` function should let the user specify which value should actually be used.

Pomax:
maybe we're approaching this from the wrong direction. We could simply update the current history state with [an object](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState()_method) that says a) which entries are loaded and b) what the scroll position is, when users scroll (using `replaceState` rather than `pushState`). That way, the browser navigation should automatically be able to hand us that object back when we navigate to a page using the "back" button.

Pomax:
Maybe it'll help if we can some clarity here: @xmatthewx is the idea that coming back to a page using the "back" button always restores where you were (e.g. entries you were seeing, and scroll position), but electing to open a new tab (say, click on featured or latest intentionally) loads a new set?

xmatthewx:
@Pomax – that is correct. We only need and want to remember scroll position for one specific scenario, when a user goes from List to Detail and back to List. It would likely feel weird in other scenarios. 

 * `Remember position` – Scenario: User opens a Detail view from somewhere in List view, then User triggers browser back from Detail to List view. >>List view displays with prior scroll position. 
 * `Everything else` – Scenario: User taps Tab (regardless of current and prior position) >> Tab loads from the top

Does that help?

mmmavis:
Cool if that's the case I think my PR should cover what @xmatthewx described above. @Pomax could you review & verify? Thanksss! 🍰

Pomax:
rereviewing

mmmavis:
@Pomax YESSS

PR #463, opened 2017-04-06
Allow manifest.json to load

### LINKED ISSUE #462: manifest.json apparently fails to load
It looks like our manifest is blocked as unsafe, at least in Chrome. Or am I mistaken? It worked previously and there haven't been any big [changes to the manifest](https://github.com/mozilla/network-pulse/blame/master/manifest.json). Maybe something happened when we went to our new domain? 

screen shot 2017-04-05 at 2 53 44 pm

As a result, when a user saves Pulse to device home screen, then opens it, web app doesn't load the start url and it doesn't load fullscreen without browser chrome.

Any insight @cadecairos @Pomax @alanmoo @mmmavis ?

mmmavis:
Fixes #462 

@flukeout this should work. help me verify? 🍺

### DIFF
--- a/js/security-headers.js
+++ b/js/security-headers.js
@@ -34,6 +34,9 @@ export default {
     ],
     frameAncestors: [
       `'none'`
+    ],
+    manifestSrc: [
+      `'self'`
     ]
   },
   reportOnly: false,


flukeout:
@mmmavis R+ tested on my android and it loads without browser chrome when saved to my homescreen and shows a splash image.

PR #464, opened 2017-04-06
quote fixes for windows

Pomax:
Windows passes single quotes as "part of the argument", so needs double quotes for these tasks to run properly


mmmavis:
@Pomax looks good to me. Do the single quotes in watch tasks work fine on Windows?

PR #476, opened 2017-04-10
Added /tags and /tags/param routes

### LINKED ISSUE #444: Add a route for tags
The API can serve up entries by Tag. As far as I can tell, we don't have a route for that right now. Let's add that. 

It's useful for program staff to have links to their stuff /tags/mozfest or /tags/global-sprint, even though we don't (yet) surface those tags in the cards.

mmmavis:
Fixes #444

### DIFF
--- a/routes.jsx
+++ b/routes.jsx
@@ -23,6 +23,10 @@ const Latest = () => {
   return <ProjectLoader />;
 };
 
+const Tag = (router) => {
+  return <ProjectLoader tag={router.params.tag} />;
+};
+
 const App = React.createClass({
   getInitialState() {
     return {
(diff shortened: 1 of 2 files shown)


routes.jsx
```diff
@@ -23,6 +23,10 @@ const Latest = () => {
   return <ProjectLoader />;
 };
 
+const Tag = (router) => {
+  return <ProjectLoader tag={router.params.tag} />;
```
alanmoo: Does this just work automagically because it hands the API "tag" as a query param?

mmmavis: yup yup

PR #494, opened 2017-04-17
More descriptive html doc title

### LINKED ISSUE #483: Add entry title to html head
For SEO and user's browser history, we should include the entry title in html head. "Foo bears – Mozilla Network Pulse"

mmmavis:
Fixes #483 

`react-helmet` doc: https://github.com/nfl/react-helmet

### DIFF
--- a/pages/entry.jsx
+++ b/pages/entry.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { browserHistory } from 'react-router';
+import { Helmet } from "react-helmet";
 import ProjectCard from '../components/project-card/project-card.jsx';
 import Service from '../js/service.js';
 import Utility from '../js/utility.js';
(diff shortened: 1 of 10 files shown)


pages/entry.jsx
```diff
@@ -50,16 +51,23 @@ export default React.createClass({
...
 
     if (this.state.dataLoaded) {
+      docTitle = `${this.state.entry.title}`;
       justPostedByUserMessage = this.state.justPostedByUser ? (<h5 className="text-center">Thanks for submitting!</h5>) : null;
       projectCard = this.state.dataLoaded ? <ProjectCard {...Utility.processEntryData(this.state.entry)} onDetailView={true} /> : null;
     }
 
     return (
       <div>
+        { docTitle &&
```
Pomax: this is basically an anti-pattern, where you're triggering core functionality as a side effect rather than clear intent.

Let's instead do something like:
```
let docTitle = null;

if (...) {
  docTitle = ...
  ...
}
if (doctitle) {
  docTitle = <Helmet><title>{ docTitle }</title></Helmet>;
}
return (
  <div>
  {docTitle}
  { just....
```

pages/issue.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
+import { Helmet } from "react-helmet";
 import IssueSelector from '../components/issue-selector/issue-selector.jsx';
 import ProjectLoader from '../components/project-loader/project-loader.jsx';
 
 export default function (props) {
-  const encodeIssueAsUri = function(issue) {
+  const getFullIssueName = function(issue) {
```
Pomax: there is no scope busting going on in this function, so we can declare it outside of the export instead of inlining it here.

We can also improve this code because we don't even need a function here:
```
const mapping = {
  "online-privacy-and-security": "Online Privacy & Security",
  "open-innovation": "Open Innovation",
  ...
}

export default function (props) {
  const issueName = mapping[props.params.issue] || ``;

  return <div>
            <Helmet><title>{issueName}</title></Helmet>
            <IssueSelector />
            <ProjectLoader issue={encodeURIComponent(issueName)} />
          </div>;
};
```

Issue #477, opened 2017-04-10
Migrate React Components to ES6 syntax


mmmavis:
`React.createClass` is being deprecated. Migrate React Components to ES6 syntax so we don't run into `react` upgrading problem in the future.

Relevant links:
- https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html
- https://toddmotto.com/react-create-class-versus-component/

Sudip7:
@mmmavis  would like to offer some contribution. May not be the complex one but smaller portions of some files.

mmmavis:
@Sudip7 cool, you can pick the ones you are comfortable tackling. Just make sure your PR references this ticket! 😃

mmmavis:
Hey @kamos1 😃 Thanks for offering help. Feel free to pick any bits you'd like to help with ES6ing. In your PR please mention this issue number. One note is that we are currently working on a new feature on a feature branch. This feature branch will likely be merged to `master` (our primary branch on this repo) soon so do expect a ton of code changes in the next few weeks.

kamos1:
Hey @mmmavis, as components are ES6'd should I be using proptypes? I've seen them used in other ES6'd components

mmmavis:
THIS IS DONE!!! Thanks everyone who contributed! 🍰 ☕️ 🍵 🍺 🍫 :shallow_pan_of_food:

PR #495, opened 2017-04-17
Migrating away from React.PropType

### LINKED ISSUE #478: Migrate from React.PropTypes to using the `prop-types` library
`React.PropTypes` is deprecated as of React v15.5. Use the `prop-types` library instead.

https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes

mmmavis:
fixes #478 & covers some of #477 

I'm still seeing a warning about `React.PropType` but I think it's coming from the modules we are using.
screen shot 2017-04-17 at 3 14 38 pm

### DIFF
--- a/components/project-card/project-card.jsx
+++ b/components/project-card/project-card.jsx
@@ -1,20 +1,20 @@
 import React from 'react';
 import ReactGA from 'react-ga';
 import { Link } from 'react-router';
+import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import moment from 'moment';
 import { getBookmarks, saveBookmarks } from '../../js/bookmarks-manager';
 
-const Details = React.createClass({
-  propTypes: {
-    createGaEventConfig: React.PropTypes.func.isRequired
-  },
+class Details extends React.Component {
   handleVisitBtnClick() {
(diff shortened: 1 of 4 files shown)


components/project-card/project-card.jsx
```diff
@@ -212,4 +212,24 @@ export default React.createClass({
...
+}
+ProjectCard.propTypes = {
+  id: PropTypes.number.isRequired,
+  creators: PropTypes.array,
+  description: PropTypes.string.isRequired,
+  featured: PropTypes.bool,
+  getInvolved: PropTypes.string,
+  getInvolvedUrl: PropTypes.string,
+  interest: PropTypes.string,
+  issues: PropTypes.arrayOf(React.PropTypes.string),
```
Pomax: so sneaky~~ see that `React.PropTypes`? that's where the warning's hiding

Sudip7: Typical copy, paste n a miss😀

mmmavis: I did search for `React.PropTypes` and got no matching result back. Weird. :nauseated_face: I should've been more careful!

Thanks for catching it @Pomax

Issue #506, opened 2017-05-02
Add a 404-a-like when accessing entries that don't exist.


Pomax:
Due to https://github.com/mozilla/network-pulse-api/pull/114 we need to update the client so that it can inform users that the entry they asked for "does not exist" when requested as `http://.../entry/12` - this can now happen when the entry's `is_approved` flag is set to False (e.g. pending moderation), in which case the entry will be excluded from API calls.

Pomax:
Rather than revealing whether an entry does not exist, or whether it is pending approval, I'd like to suggest we make this say "This entry is not current available". Does that seem reasonable @alanmoo @xmatthewx?

PR #509, opened 2017-05-02
special route for posts pending approval

Pomax:
closes https://github.com/mozilla/network-pulse/issues/504, closes https://github.com/mozilla/network-pulse/issues/507

This adds in a new route for submittion entries that are not immediately set to `is_approved`.
- user hits up "add" page
- user submits entry data
- if the entry is immediately `is_approved` (e.g. this is a staff post) then the user is redirect to their entry's URL (same as before)
- if the entry is **not** immediate `is_approved` (e.g. normal user) then they are sent to `/submitted` and presented with a thank you screen and a link to where their entry will exist once we approve it.

PR #510, opened 2017-05-02
Give an "entry unavailable" notice on 404 and pending approval

Pomax:
Closes https://github.com/mozilla/network-pulse/issues/506

This change adds some reporting to the `/entry` page so that when an entry is a 404, or it's filtered as a 404 because it's pending approval, the page is not just a blank page, but informs the user that this entry is not currently available.

This PR is built on top of #509


alanmoo:
Just make the changes to #509 first and rebase

Issue #512, opened 2017-05-02
JS errors while having, as well as on the favs page


Pomax:
there is a `this is undefined` error when faving, and the `/favs` page has an error about fetchData being missing. Both these errors mean that faving is completely broken.

PR #513, opened 2017-05-02
fix event scoping and nonexistent function calls

Pomax:
fixes https://github.com/mozilla/network-pulse/issues/512 by resolving a scoping issue where ES6 syntax is used for JSX but the event handling does not use `evt => this.doThing(evt)` to preserve `this` scoping.  It also removes a reference to `this.fetchData()` in the bookmark page, which didn't exist and was probably a copy-paste leftover.

### DIFF
--- a/pages/bookmarks.jsx
+++ b/pages/bookmarks.jsx
@@ -30,8 +30,6 @@ export default React.createClass({
       if (pageSettings.shouldRestore) {
         // restore state back to what is stored in pageSettings
         this.setState(pageSettings.currentList);
-      } else {
-        this.fetchData();
       }
     });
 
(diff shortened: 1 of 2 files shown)


pages/bookmarks.jsx
```diff
@@ -30,8 +30,6 @@ export default React.createClass({
       if (pageSettings.shouldRestore) {
         // restore state back to what is stored in pageSettings
         this.setState(pageSettings.currentList);
-      } else {
```
alanmoo: Why'd this go away?

Pomax: this object has no `fetchData` function - instead it should just do "whatever the code does normally" when not explicitly instructed to use the cached results.

PR #536, opened 2017-05-16
More responsive cards. CSS cleanup to project cards and nav bar.

### LINKED ISSUE #491: Make cards more responsive
from QA ticket https://github.com/mozilla/network-pulse/issues/419

### Action
View pulse at different screen widths

#### 1) Thumbnail grid view
- Cards are squished very thin when they don't need to be (see screen cap from product)
- we should switch to 2-per-row below ~992px
- recommend a minimum card width ~320px
screen shot 2017-04-17 at 9 37 35 am

#### 2) Detail view
- Individual card has tons of margin at browser widths below 992px  (see screen cap from product)
- cards should keep the wider style for longer
screen shot 2017-04-17 at 9 40 51 am

mmmavis:
Fixes #491... plus CSS cleanup to project cards and nav bar.

@alanmoo sorry for the not-so-small PR. I couldn't help but clean up some custom CSS code with Bootstrap classes. I've QAed them on Chrome, Firefox, and Safari and they all seem good to me. The only thing that's different from what we had before is the gutter width between project cards now is wider. Once we get the changes up on staging I will ask feedback/suggestion from designers and we can tweak styling from there.

### DIFF
--- a/components/project-card/project-card.scss
+++ b/components/project-card/project-card.scss
@@ -1,17 +1,11 @@
 .project-card {
-  box-sizing: border-box;
-  width: 100%;
+  margin-bottom: 20px;
   background: $white;
   overflow: hidden;
-  margin-left: $project-card-side-margin;
-  margin-right: $project-card-side-margin;
-  margin-bottom: 20px;
-
   /* optimize swipe with class and transform. */
   transition: left 0.25s, height 0.25s, padding 0.25s, margin 0.25s;
 
   &.detail-view {
-    width: 100%;
     margin: 0 auto;
 
     h2 {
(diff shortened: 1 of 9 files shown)


components/project-card/project-card.scss
```diff
@@ -146,10 +140,8 @@
 // >= 768px
 @media screen and (min-width: $bp-md) {
   .project-card {
-    @include project-card-width(3);
-
     &.detail-view {
-      @include project-card-width(2);
+      max-width: 600px;
```
alanmoo: Any reason you wouldn't use a column mixin (or class on the element itself) so we're consistently using the grid? (Thinking `col-12 col-md-8` or something like that comes close to 600px)

mmmavis: Nice catch! I've updated the PR.

mmmavis:
🎉

Issue #489, opened 2017-04-17
Reduce title size and length 


kristinashu:
Some titles are long: (screen cap from production)
<img width="1433" alt="screen shot 2017-04-17 at 8 44 47 am" src="https://cloud.githubusercontent.com/assets/1068783/25094252/2312f49c-234b-11e7-948d-d22c0aeaa272.png">

We should reduce font size slightly (from 30px to 26px).

@xmatthewx do you think we need to add a character count to the title?

xmatthewx:
oye. i think there is a character count, but maybe it should be even shorter. smaller type might help. 

there is an issue to update some copy on the form for legal. i added a comment about titles, but it'll only help so much. https://github.com/mozilla/network-pulse/issues/484#issuecomment-294195199

we reserve the right to edit entries. i think if we do a round of cleanup, people will hopefully write titles  similar to what they see. 

or... maybe we should move away from titles. maybe titles are a relic of the past. i'm an old person so i'm not sure. should we let people tweet, merging title and description?

kristinashu:
Whaaat?!? Long live titles! 

Things to do:
- [ ] reduce font size in title
- [ ] manually clean up existing titles
- [ ] reduce character count from 140 to 80
- [ ] update language - #484 
- [ ] inform users about character count - #493

kristinashu:
@xmatthewx can we put this one into our next heartbeat?

PR #543, opened 2017-05-18
Tag suggestions on form

### LINKED ISSUE #482: Add autocomplete for tags on input form
Autocomplete can help users avoid typos and use consistent tags, so that people who `games` can also find a good `game`. 

Note: We should convert to lowercase

Step 1 – a designer and dev should collaborate to define requirements, gather benchmarks, and scope the simplest useful solution. Interaction will deeply intertwine both disciplines.

Step 2 – I recommend we prototype rather than comp-then-dev.

mmmavis:
Fixes #482
Also addresses #545

Design and styling can be improved but let's just get the functionality live first! 😉

### DIFF
--- a/components/project-card/project-card.jsx
+++ b/components/project-card/project-card.jsx
@@ -203,6 +203,10 @@ class ProjectCard extends React.Component {
     );
   }
 
+  renderDescription() {
+    return this.props.description.split(`\n`).map(paragraph => <p>{paragraph}</p>);
+  }
+
   renderIssuesAndTags() {
     if (!this.props.onDetailView) return null;
 
@@ -243,7 +247,7 @@ class ProjectCard extends React.Component {
                 {this.renderCreatorInfo()}
                 {this.renderTimePosted()}
               </div>
-              <p className="description">{this.props.description}</p>
+              <div className="description">{this.renderDescription()}</div>
               <Details {...this.props} createGaEventConfig={this.createGaEventConfig} />
             </div>
             <div className="fade-overlay"></div>
--- a/pages/add/form/detail-info-fields.js
+++ b/pages/add/form/detail-info-fields.js
@@ -14,7 +14,6 @@ let Tags = React.createClass({
     Service.tags
       .get()
       .then((data) => {
-        console.log(data);
         let suggestions = data.map((tag,i) => {
           return { id: i+1, name: tag };
         });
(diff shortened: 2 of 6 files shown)


pages/add/form/detail-info-fields.js
```diff
@@ -1,6 +1,105 @@
...
+  render: function() {
+
+    return <ReactTags
+              tags={this.state.tags}
+              suggestions={this.getFilteredSuggestions()}
+              allowNew={true}
+              handleDelete={this.handleDelete}
+              handleAddition={this.handleAddition}
+              handleDrag={this.handleDrag}
+              onChange={this.onChange}
```
Pomax: you need an `onChange` function still, which then "falls through" to `this.props.onChange` (which is handled by the form).
```
  onChange: function(evt) {
    // do a thing!
    this.setState({
      // set a thing!
    }, () => {
      let values = this.state.tags.slice();
      this.props.onChange(evt, values);
    });
  },
```

pages/add/form/detail-info-fields.js
```diff
@@ -1,6 +1,105 @@
...
+                searchInput: `tag-input-field`,
+                suggestions: `suggestions p-2`,
+                suggestionActive: 'is-active',
+                suggestionDisabled: 'is-disabled'
+              }}
+            />;
+  }
+});
+
+let Test = React.createClass({
```
Pomax: this works as long it uses `this.props.onChange` \o/

components/project-card/project-card.jsx
```diff
@@ -203,6 +203,10 @@ class ProjectCard extends React.Component {
     );
   }
 
+  renderDescription() {
+    return this.props.description.split(`\n`).map(paragraph => <p>{paragraph}</p>);
```
mmmavis: @Pomax Is `\n` the right delimiter? `\n\r`?

Pomax: since the content came from javascript, `\n` should be good enough. If you're not sure where it came from, then it's typically `.split(/\r?\n/)` to capture the possible carriage return.

Pomax:
One thing we probably want to do is make sure that if there's a tag match for some specific capitalisation, we substitute. So if there's a tag DIY and the user types "diy[enter]" we substitute with DIY.

mmmavis:
@Pomax great suggestion! Should be a follow-up after https://github.com/mozilla/network-pulse/issues/489 is done!

mmmavis:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

Issue #547, opened 2017-05-19
Correct Sign In messages


xmatthewx:
Now that we've opened things to non-staff with google login, we should tweak our help text.

 - [ ] Remove "Are you Mozilla Staff?" from Favs view
 - [ ] Remove "Please note, only Mozilla..." from "Add" view

"Add" view says sign in, which can be slightly confusing if I don't already have an account. Seems like "Sign in with Google" is the most clear and accurate button label. Do you agree @kristinashu? 

 - [ ] Change button to "Sign in with Google"

kristinashu:
Yes, sounds good!

alanmoo:
I updated some of the wording and it's on staging, but since we've got actual directive another PR would be useful

xmatthewx:
@hannahkane can you check in on this tomorrow?

alanmoo:
Closed by #553

Issue #454, opened 2017-04-02
Different message on Favs tab


Sudip7:
Hi team,

I am not sure whether it's really a issue or not but I have noticed this more than one time during development. I also don't know precisely how to mock/reproduce this issue. I tried few way but none is working. For your reference adding a screen shot.

The piece of code this message belongs to present in bookmarks.jsx. But there was no error in the terminal , I forgot to test browser terminal though. 

`getFailurePrompt() {
    return (<p>Only Mozilla staff can login now as we test this new platform. Meanwhile, your favorited projects can still be remembered on this device.</p>);
  }`

Please look into it. 

![favfirsttime](https://cloud.githubusercontent.com/assets/6715793/24588473/b88566f6-177d-11e7-80c1-76ad1f016340.jpeg)

@mmmavis @Pomax  any ideas!!!

mmmavis:
@Sudip7 you mentioned you have been seeing different messages when visiting the Favs tab. The Screencap counts as one, what about other messages you have seen? Do you remember what they were?

mmmavis:
Messages on that Favs tab varies when your login state changes. There are 3 cases:

## Before a user attempts to log in

<img width="1126" alt="screen shot 2017-04-03 at 1 37 22 pm" src="https://cloud.githubusercontent.com/assets/2896608/24630772/c77d4e76-1872-11e7-8fa3-e390c7417b06.png">

## User has successfully logged in (currently only Mozilla Staff can log in)
<img width="1432" alt="screen shot 2017-04-03 at 1 39 22 pm" src="https://cloud.githubusercontent.com/assets/2896608/24630831/fadcd5e8-1872-11e7-9eba-ea994e40d1cc.png">

## User is not allowed to log in (when non Mozilla Staff tries to log in)
As seen in your screencap

Sudip7:
@mmmavis: the other screen is the regular one, that contains the sign in link. I believe this is the default behaviour. Btw how u distinguish Mozilla and non-mozilla users???

I found this particular screen few times so thought to enquire about it.

xmatthewx:
I believe this message should only appear on the Favs screen. Login is currently limited to staff, based on the domain of the email address user submits to login.

Is there a bug? Or are you just asking about this behavior? We plan to gradually open this up to more users after we have a chance to test and build a bit more.

mmmavis:
[...]

PR #553, opened 2017-05-23
sign in with Google

Sudip7:
#547 : sign in with Google feature added

### DIFF
--- a/pages/bookmarks.jsx
+++ b/pages/bookmarks.jsx
@@ -49,10 +49,10 @@ export default React.createClass({
     return(<p>Hi {user.username}! <button onClick={this.handleLogOutBtnClick} className="btn btn-link inline-link">Log out</button>.</p>);
   },
   getFailurePrompt() {
-    return (<p>Only Mozilla staff can login now as we test this new platform. Meanwhile, your favorited projects can still be remembered on this device.</p>);
+    return (<p>Something went wrong! Please refresh and try again.</p>);
   },
   getAnonymousContent() {
-    return (<p>Are you Mozilla Staff? <button className="btn btn-link inline-link" onClick={this.handleSignInBtnClick}>Sign in</button>.</p>);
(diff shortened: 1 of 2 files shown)


pages/bookmarks.jsx
```diff
@@ -49,10 +49,10 @@ export default React.createClass({
     return(<p>Hi {user.username}! <button onClick={this.handleLogOutBtnClick} className="btn btn-link inline-link">Log out</button>.</p>);
   },
   getFailurePrompt() {
-    return (<p>Only Mozilla staff can login now as we test this new platform. Meanwhile, your favorited projects can still be remembered on this device.</p>);
+    return (<p>Something went wrong! Please refresh and try again.</p>);
```
mmmavis: @Pomax since now all people with a Google account can sign in, will user who **does not** have a Google account ever be redirected back [to the original_url that we passed as the sign in url query](https://github.com/mozilla/network-pulse-api#get-loginoriginal_urlurl)? I'm assuming he/she will be "stuck" on the Google Sign In page after clicking the "sign in" link on Pulse. If that's the case we might not need this `getFailurePrompt()` and other relevant code anymore?

Sudip7: Is there any plan in the pipeline for sign up if they don't have an account(here Google). I added that message as we have a bug #454 but again if there is a provision to add sign up options in future we may add some buttons. But I still believe there should some sort message in case of a loading error.

@mmmavis @Pomax

xmatthewx:
@mmmavis @Pomax @alanmoo – let's prioritize getting this wrapped up today. There will a lot of new visitors on Pulse today.

pages/bookmarks.jsx
xmatthewx: @Sudip7 – Right now we only have google login and we don't have specific plans to change that at this time. No need to worry about that now. Thanks!

Issue #517, opened 2017-05-03
Add number of results for each issue tag


kristinashu:
@xmatthewx would you be up for adding something like this? This will be helpful feedback for users when one post appears at the top of several issues.

 https://redpen.io/qf3e141d69fb01cf4c

![issues results 2x](https://cloud.githubusercontent.com/assets/1068783/25675078/06771508-2ff2-11e7-9081-d050d64b234f.png)

xmatthewx:
Yep. Great idea. Useful for search as well.

Sudip7:
is this enhancement applicable for issues tab or for all the relevant tabs. Please clarify. @xmatthewx @kristinashu

kristinashu:
Only in the Issues tab, thanks!

Note that search results already does this ([example](https://www.mozillapulse.org/search?keyword=design)), please style it the same.

Sudip7:
Thanks @kristinashu . I added some code. Please, suggest should I add additional filters in order to restrict that to issue page or keep as it is.

![image](https://cloud.githubusercontent.com/assets/6715793/26260425/c8b58e7e-3cea-11e7-90ec-b042380a1e81.png)

![image](https://cloud.githubusercontent.com/assets/6715793/26260460/f5f3496c-3cea-11e7-9aef-7f9f0fcdc7a6.png)

![image](https://cloud.githubusercontent.com/assets/6715793/26260582/5ebfc024-3ceb-11e7-9e20-c543d413fa0b.png)


![image](https://cloud.githubusercontent.com/assets/6715793/26260524/2a3e8218-3ceb-11e7-89c0-9d471ce3c4f6.png)


@xmatthewx  @mmmavis  please suggest.

kristinashu:
Yes, please hind on Featured and Latest pages. Thanks!

xmatthewx:
👍  Yes. Please hide everywhere. Everywhere but individual issue pages. It could also show on individual tag views, for example: [/tags/mozfest](https://www.mozillapulse.org/tags/mozfest). But it shouldn't appear anywhere else.

Sudip7:
I have added following code: 

reIssueCountResult() {
    if(this.props.latest) return null; 
    if(this.props.featured || this.props.search || this.state.loadingData) return null;
let total = this.state.totalMatched,
  plural = (total === 0 || total > 1), // because "0 results"
  issueCountResult = `${total} result${plural ? `s` : ``} found`;

return <p>{issueCountResult}</p>;
  },

[...]

PR #554, opened 2017-05-23
adding issue counter

Sudip7:
#517 : Issue counter added.

### DIFF
--- a/components/project-loader/project-loader.jsx
+++ b/components/project-loader/project-loader.jsx
@@ -130,12 +130,23 @@ export default React.createClass({
 
     return <p>{searchResultNotice}</p>;
   },
+  renderIssueCountResult() {
+    if(this.props.latest) return null; 
+    if(this.props.featured || this.props.search || this.state.loadingData) return null;
+    
+    let total = this.state.totalMatched,
+      plural = (total === 0 || total > 1), // because "0 results"
+      issueCountResult = `${total} result${plural ? `s` : ``} found`;
+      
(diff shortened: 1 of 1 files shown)


components/project-loader/project-loader.jsx
```diff
@@ -130,12 +130,23 @@ export default React.createClass({
...
+  renderIssueCountResult() {
+    if(this.props.latest) return null; 
+    if(this.props.featured || this.props.search || this.state.loadingData) return null;
+    
+    let total = this.state.totalMatched,
+      plural = (total === 0 || total > 1), // because "0 results"
+      issueCountResult = `${total} result${plural ? `s` : ``} found`;
+      
+    return <p>{issueCountResult}</p>;
+  },
```
mmmavis: from https://github.com/mozilla/network-pulse/issues/517#issuecomment-302813765 we are asked to display The counter on individual issue pages and tag pages **only**.

I would change the above code to something like

```jsx
renderEntryCounter() {
  if (this.state.loadingData) return null;
  if (!this.props.issue && !this.props.tag) return null;

  return <p>{this.state.totalMatched} result{this.state.totalMatched > 0 ? `s` : ``}</p>;
}
```

mmmavis: Note that we should change name of the method to `renderEntryCounter` since we are displaying number of the matched entries and not the issue count.

Sudip7:
@mmmavis , please review

Issue #441, opened 2017-03-22
More friendly form errors (not visible from below)


xmatthewx:
While posting a new entry today, Submit didn't work and it wasn't clear why. 

As you'll see in the screenshot below, the button is enabled, but a click yields no action. I assumed there was a bug. I scrolled up to take a screenshot, and only then noticed the Error message on one of the fields. 

To make this more friendly, we should do something. Maybe scroll up to error on submit attempt. Maybe show an error by button that has a generic message to correct an error up above. 

<img width="1116" alt="screen shot 2017-03-22 at 11 52 57 am" src="https://cloud.githubusercontent.com/assets/53411/24207159/95c35b4e-0ef6-11e7-864b-49c578961faf.png">

mmmavis:
@xmatthewx Did you fill out the required fields(title and url)? There could be inline error messages that you have missed. But yea I agree we can have a more friendly UX here.

xmatthewx:
Yep. It's not a form bug. I had too many characters on a field up above. I found it because I knew that might be the case. Many users could give up. 

I recommend that we keep Submit gray and disabled when errors are present. Add helper text next to it. Something like so...

<img width="562" alt="screen shot 2017-03-24 at 12 43 50 pm" src="https://cloud.githubusercontent.com/assets/53411/24304413/9e3bde78-108f-11e7-9719-8b46ee4e3259.png">

kristinashu:
Can we do inline validation for required fields?

kristinashu:
Opened a ticket for the title word count error https://github.com/mozilla/network-pulse/issues/493

xmatthewx:
Another user impacted by this. We need to fix this. At least adding at a simple message near button telling people to make corrections above. 

cc @alanmoo

alanmoo:
I've filed the above PR as a quick low-hanging fruit fix. The error message appears, but I didn't change any behavior of the button, as that would have required deeper investigation of the event propagation/validation checking.

mmmavis:
Alan's fix is on production now: https://www.mozillapulse.org/add

xmatthewx:
good enough!

PR #560, opened 2017-05-30
Add an error message for form validation

alanmoo:
Addresses #441. There might be more work to do there, but this was a super quick fix.

### DIFF
--- a/pages/add/add.jsx
+++ b/pages/add/add.jsx
@@ -20,7 +20,8 @@ export default React.createClass({
       formValues: {},
       authError: false,
       serverError: false,
-      submitting: false
+      submitting: false,
+      formInvalid: false
     };
   },
   componentDidMount() {
@@ -72,10 +73,12 @@ export default React.createClass({
         }
 
         if (!basicFormIsValid || !detailFormIsValid) {
+          this.setState({formInvalid: true});
           return;
         }
 
         this.setState({
+          formInvalid: false,
           submitting: true
         }, () => this.postEntry(this.state.formValues));
       });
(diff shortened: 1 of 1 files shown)


mmmavis:
@alanmoo I'm not sure about this one. I actually have tried having this as a quick fix before but it didn't feel right to me - even after the user has fixed all the validation errors that "Something isn't right. Check your info above." message is still there. This is false information which I think it's worse than our original issue (poor UX)

screen shot 2017-06-01 at 12 19 54 pm

I'll spend some time later today and see if I can figure out a fix that doesn't involve changing/expanding the `react-formbuilder` library codebase.

/cc @xmatthewx @Pomax

xmatthewx:
UX-wise, you can probably clear "something isn't right..." if a user makes any change, or puts focus on any input. i don't think you need lots of intelligence. bring it back if there's a new error when they try again. not perfect, but mvp-enough?

mmmavis:
@alanmoo do you wanna update this PR yourself and tag me for review or the other way around?

alanmoo:
It would be helpful if you could take it and I'll review

mmmavis:
@alanmoo I updated this PR but couldn't set you as reviewer since this is technically your PR 😂 Can you review?

pages/add/add.jsx
```diff
@@ -20,7 +20,8 @@ export default React.createClass({
       formValues: {},
       authError: false,
       serverError: false,
-      submitting: false
+      submitting: false,
+      showFormInvalidNotice: false
```
mmmavis: I renamed this from `formInvalid` to `showFormInvalidNotice` since form is still invalid at some point... it's just we decide to not show the message to user.

alanmoo:
I like that change! Looks good to me if everything else looks good to you. You should be able to approve it so merging can happen, right?

Issue #562, opened 2017-05-31
New tags aren't added to an entry


alanmoo:
Noticed this on staging, haven't confirmed on production but not sure why it would differ:

1. Create a new entry (entry name, link) and fill in an autocompleted tag.
2. After the tag appears in its grey box, type in something else, which should become a new tag for the entry. 
3. Click save
4. Notice that the text entered after the existing tag isn't attached to the entry. It disappeared.

mmmavis:
I can't reproduce this. My "mavistesting" tag shows up fine on staging.

<img width="1133" alt="screen shot 2017-06-01 at 3 11 11 pm" src="https://cloud.githubusercontent.com/assets/2896608/26703038/99da8446-46dc-11e7-9dcd-1edbe9797f1a.png">

alanmoo:
Were you pressing "enter" after typing in the new tag? If you do that it works properly, but it was definitely not obvious to me.

mmmavis:
@alanmoo right, I did press the "enter" key. To me using "enter" as delimiter has been pretty intuitive but as you said it might not be obvious to other people.

PR #565, opened 2017-06-01
force delimiting on comma

Pomax:
fixes https://github.com/mozilla/network-pulse/issues/562, the ninja hack way.

Ideally, this hack can be taken out again after https://github.com/i-like-robots/react-tags/issues/76 is addressed.

### DIFF
--- a/pages/add/form/detail-info-fields.js
+++ b/pages/add/form/detail-info-fields.js
@@ -64,7 +64,7 @@ let Tags = React.createClass({
     this.updateTags(tags);
   },
   handleAddition: function(tag) {
-    tag.name.trim();
+    tag.name = tag.name.trim();
     const tags = [].concat(this.state.tags, tag);
     this.updateTags(tags);
   },
--- a/pages/add/form/detail-info-fields.js
+++ b/pages/add/form/detail-info-fields.js
@@ -29,6 +29,9 @@ let Tags = React.createClass({
     // The ReactTags component for some reason does not
     // do tag seperation on anything except "enter" which is
     // not an intuitive key when you're on a single-line input.
+    // We are adding a hack here to force delimiting on comma.
+    // Ideally, this hack can be taken out again after
+    // https://github.com/i-like-robots/react-tags/issues/76 is addressed.
     try {
       const reactTags = this.reactTags;
       const trueInput = reactTags.input.input;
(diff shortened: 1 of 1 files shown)


alanmoo:
Is there a spot we can stick a `trim` so that if a user types `comma space` the new tag doesn't get saved with a leading space?

pages/add/form/detail-info-fields.js
```diff
@@ -22,6 +22,32 @@ let Tags = React.createClass({
...
+      const trueInput = reactTags.input.input;
+      const forceKey = reactTags.handleKeyDown.bind(reactTags);
+
+      trueInput.addEventListener("keyup", evt => {
+        // did we type a comma?
+        const key = evt.key;
+        if (key === ',') {
+          // remove the comma and type an enter instead
+          const { query, selectedIndex } = reactTags.state;
+          reactTags.state.query = query.substring(0, query.length-1).trim();
```
alanmoo: This isn't the right spot for this, doesn't get triggered on the final tag if the user separates with tab/return

alanmoo: After line 63 add  `tag.name.trim();`

mmmavis: I didn't catch ^ and I just R+'ed this PR. Let me take a closer look again.

mmmavis: like below?

```jsx
updateTags: function(tags) {
    this.setState({ tags }, () => {
      const tagNames = this.state.tags.slice().map(tagObj => tagObj.name.trim());
      this.props.onChange(null,tagNames);
    });
  },
```

pages/add/form/detail-info-fields.js
```diff
@@ -35,7 +64,7 @@ let Tags = React.createClass({
     this.updateTags(tags);
   },
   handleAddition: function(tag) {
-    const tags = [].concat(this.state.tags, tag);
+    const tags = [].concat(this.state.tags, tag.trim());
```
mmmavis: 👍

alanmoo: Nope, needs to be `tag.name.trim()` on the line before. `tag` is an object, so we can't directly trim it, and we can't pass `concat` just the name string.

mmmavis:
Got okay from both @alanmoo and @Pomax . I'll merge this on Monday since it's almost EOD Friday 🏖

PR #575, opened 2017-06-19
Moderation page

### LINKED ISSUE #541: Build moderation front-end
Moderating many projects would benefit from a frontend view.

Why not just use Django? 
 - it's unfamiliar. and we'll want many moderators with minimal training required
 - Django doesn't currently provide a way to browse by moderation state
 - Django doesn't currently make it easy to read a project. See below. 

screen shot 2017-05-17 at 1 12 20 pm

### Frontend specs

 - List view MVP
   - show non-approved projects at /pending/ or similar
   - indicate state on each card (pending, in-review, declined)
 - Detail view MVP
   - a drop-down to change moderation state (see #522)

### Nice to have (not likely for first build) 
 - List view
   - way to filter state (pending, in-review)
   - a way to filter by tag or search (e.g. see all global sprint)

Any other requirements?
@hannahkane @kristinashu

mmmavis:
Fixes #541 

Features added in this PR:

- a `/moderation` page (404 for non-moderator)
- on moderation page:
  - by default this page shows no entries until you type in some keywords (same behaviour as we have on the /search page)
  - search entries by keyword
  - filter entries by moderation state
  - update entry's moderation state
  - `keyword` and `moderationstate` should appear as url query


js/security-headers.js
```diff
@@ -28,7 +29,7 @@ export default {
     ],
     connectSrc: [
       `'self'`,
-      env.PULSE_API || `https://network-pulse-api-staging.herokuapp.com/`
+      url.parse(env.PULSE_API).host || `https://network-pulse-api-staging.herokuapp.com/`
```
mmmavis: @Pomax should this be `https://network-pulse-api-staging.herokuapp.com/` or `https://network-pulse-api-staging.herokuapp.com/api/pulse/`?

package.json
```diff
@@ -9,11 +9,11 @@
     "build": "run-s bootstrap build:**",
     "build:config": "node js/build-env-config.js > config/env.generated.json",
     "build:client": "run-s build-client:**",
-    "build-client:js": "webpack --config webpack.config.js --optimize-minimize --display-error-details --colors",
+    "build-client:js": "webpack --config webpack.config.js --display-error-details --colors",
     "build-client:scss": "node-sass scss/main.scss dist/css/main.css",
     "build-client:prefix:main": "postcss --use autoprefixer -o dist/css/main.css dist/css/main.css",
     "build-client:static": "run-s copy:**",
-    "build:server": "webpack --config webpack.server.config.js --optimize-minimize --display-error-details --colors",
```
mmmavis: we only need to minimize things for for `production` and this is handled in the webpack config files.

js/security-headers.js
mmmavis: oh nvm i think this is good

components/project-card/project-card.jsx
```diff
@@ -169,14 +252,14 @@ class ProjectCard extends React.Component {
...
-    if (this.props.onDetailView) {
+    if (this.props.onDetailView || this.props.onModerationMode) {
       let twitterUrl = `https://twitter.com/intent/tweet?text=${encodeURIComponent(this.props.title)}&url=${encodeURIComponent(window.location.href)}`;
       actionPanel = (
         <div className="d-flex share">
           <a href={twitterUrl} onClick={evt => this.handleTwitterShareClick(evt) } className="btn twitter-share d-inline-block align-self-center mr-3"></a>
           <div className="reveal-url">
             <a className="btn" onClick={evt => this.handleShareBtnClick(evt)} ref={(btn) => { this.shareBtn = btn; }}></a>
-            <input creadOnly type="text" ref={(input) => { this.urlToShare = input; }} className="form-control px-2" />
+            <input readOnly type="text" ref={(input) => { this.urlToShare = input; }} className="form-control px-2" />
```
Pomax: =P

components/project-card/project-card.jsx
```diff
@@ -217,11 +300,11 @@ class ProjectCard extends React.Component {
   }
 
   renderDescription() {
-    return this.props.description.split(`\n`).map(paragraph => <p>{paragraph}</p>);
+    return this.props.description.split(`\n`).map((paragraph, i) => <p key={i}>{paragraph}</p>);
```
Pomax: Don't use list position as a key. Just use `key={paragraph}` so that reordering doesn't incur heavy DOM modification

components/project-card/project-card.jsx
```diff
@@ -294,7 +385,8 @@ ProjectCard.propTypes = {
 };
 
 ProjectCard.defaultProps = {
-  onDetailView: false
+  onDetailView: false,
+  moderationState: `` // id of the moderation
```
Pomax: if this is an id, having it be a string by default is kind of weird. Why not `false` or `undefined`?

components/project-loader/project-loader.jsx
```diff
@@ -54,6 +55,11 @@ export default React.createClass({
       return Object.assign(combinedParams, { page: 1 });
     }
 
+    if (combinedParams.moderationState) {
+      combinedParams.moderationstate = combinedParams.moderationState.label;
```
Pomax: This is a really bad idea. At the very least, rename this from `combinedParams.moderationstate` to `combinedParams.moderationStateLabel` so that it doesn't look like you're saying `x.a = x.a.b`. To anyone with even the mildest form of dyslexia, including someone who's not had their morning coffee yet, that case difference isn't there.

components/project-loader/project-loader.jsx
```diff
@@ -117,37 +123,30 @@ export default React.createClass({
...
+    return <div><p>Discover & collaborate on projects for a healthy internet. <a href="https://www.mozillapulse.org/entry/120">Learn more</a>.</p></div>;
   },
   renderEntryCounter() {
     if (this.state.loadingData) return null;
-    if (!this.props.issue && !this.props.tag) return null;
-    
-    return <p>{this.state.totalMatched} result{this.state.totalMatched > 0 ? `s` : ``} found</p>;
+    if (!this.props.search && !this.props.issue && !this.props.tag) return null;
+
+    let counterText = `${this.state.totalMatched} result${this.state.totalMatched > 0 ? `s` : ``} found`;
```
Pomax: This needs to be `this.state.totalMatched === 1? '' : 's'`:

- 0 result**s** found
- 1 result found
- 2 result**s** found.

pages/search/search.jsx
```diff
@@ -58,22 +80,77 @@ export default React.createClass({
       // In the meanwhile, we have to rely on document.querySelector(`#search-box`) to trigger input's focus() function
       document.querySelector(`#search-box`).focus();
     });
-  },
+  }
+
+  renderSearchBar() {
+    return <div className={classNames({activated: true, 'search-bar': true, 'mb-2': this.props.moderation})}>
+            <DebounceInput id="search-box"
```
Pomax: why not indent it like you indent https://github.com/mozilla/network-pulse/pull/575/files#diff-0634d7511c187586b619180db9c55f90R121 ? `id="..."` on the next line, indented, and `/>` on its own line at the same indent as the `<DebounceInput` opening

components/project-card/project-card.jsx
```diff
@@ -1,12 +1,95 @@
...
 import { Link } from 'react-router';
+import Select from 'react-select';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import moment from 'moment';
 import { getBookmarks, saveBookmarks } from '../../js/bookmarks-manager';
+import Service from '../../js/service.js';
 import Utility from '../../js/utility.js';
 
+class ModerationPanel extends React.Component {
```
Pomax: Can we put these things in their own files? It's *much* easier to work with if components are in files named after the classes defined in them.

components/project-card/project-card.jsx
```diff
@@ -1,12 +1,95 @@
...
+  }
+}
+
+ModerationPanel.defaultProps = {
+  id: ``,
+  moderationState: ``
+};
+
+
 class Details extends React.Component {
```
Pomax: same requests here I guess.

Pomax:
I don't see the moderation page either in `pages` or in `components`, which makes it very hard to locate the files involved unless you already know the codebase. Can we improve this a little?

routes.jsx
```diff
@@ -38,6 +39,35 @@ const Tag = (router) => {
         </div>;
 };
 
+
+class Moderation extends React.Component {
```
Pomax: please move this into its own file as `/pages/moderation.jsx`

Pomax:
I tried running this locally but I do not see it present me with the actual list of entries when I go to the moderation tab. We probably need to change how we're presenting/structuring that moderation tab: right now this is an overload on the `Search` tab, but that's not in line with what moderation means. This should be its own page with as primary role "showing all entries that need to be moderated" based on the selected moderation state, maybe with an optional search to refine inside those sets, but 99% of the time that search bar will not be usefulbecause you're not looking for specific entries, you're looking for "all entries with moderation state X" so that you can run through them and change moderation states.

Pomax:
After talking with hannah, the idea is to start with:
- de-emphasized search bar
- primary UI is the pull down menu for picking moderation state, with "Pending" automatically selected (this is the first moderation state in the list of states, so we can select based on that rather than hardcoding the string `Pending`)
- show all entries for any chosen moderation state (but probably with the same paging that we have on the `latest` page)
- only refine the results when the search bar is used.

js/service.js
```diff
@@ -121,15 +121,16 @@ function callURL(route) {
 }
 
 /**
- * Make an POST XHR request and return a promise to resolve it.
+ * Make a POST or PUT XHR request and return a promise to resolve it.
  * @param  {Object} params A key-value object to be posted
  * @returns {Promise} A promise to resolve an XHR request
  */
-function postEntry(entryData) {
+function updateEntry(requestType = ``, endpointRoute, entryData) {
```
Pomax: we'll need to update the code in this function so that it considers all HTTP `20x` as "things worked" - with the moderation PUT operation returning a 204 code ("everything okay, there just isn't any data to send back to you") the current code treats that 204 as "not 200" and so it mistakenly identifies the code as an error

components/project-card/project-card.jsx
mmmavis: Oh yea I thought about that but wasn't sure if having a long key (`paragraph`) would be problematic. Will update this line.

components/project-loader/project-loader.jsx
mmmavis: Ummm key in `combinedParams` will later be passed as query params to the `/entry/?` call. `moderationstate` is the query param the API understands so that's why I didn't give it another (better) name. If it helps I will drop an inline note there.

Pomax:
This is looking pretty good from a code and flow perspective! We should probably get @kristinashu to give this a quick look over for the design perspective: I'm not sure the separator is necessary here.

screenshot 215


screenshot 216

kristinashu:
Yeah looks good! You could remove the divider or update it so it's just one medium grey line.

mmmavis:
I changed the border styling to be one solid grey line!

mmmavis:
Got a R+ from @Pomax yay! Merging!

PR #581, opened 2017-06-20
Removed interim fix for tag field

### LINKED ISSUE #568: remove the "delimit on comma" hack
We introduced a workaround for `react-tags` in https://github.com/mozilla/network-pulse/pull/565 that we should be removing again as early as possible, contingent on https://github.com/i-like-robots/react-tags/pull/80 landing.

Once it has, we can update our  version of `react-tags` and use the new `delimiters` property to pass in the keyCode for comma (`delimiters={[13, 44]}`)

mmmavis:
Fixes #568 

---

Note: confirm if we want to push for oneWord tags (no spaces) or multi-word tags (with spaces) are fine

### DIFF
--- a/pages/add/form/detail-info-fields.js
+++ b/pages/add/form/detail-info-fields.js
@@ -22,35 +22,6 @@ let Tags = React.createClass({
       .catch((reason) => {
         console.error(reason);
       });
-
-      this.shimTagDelimiter();
-  },
-  shimTagDelimiter: function() {
-    // The ReactTags component for some reason does not
-    // do tag seperation on anything except "enter" which is
-    // not an intuitive key when you're on a single-line input.
-    // We are adding a hack here to force delimiting on comma.
-    // Ideally, this hack can be taken out again after
-    // https://github.com/i-like-robots/react-tags/issues/76 is addressed.
-    try {
-      const reactTags = this.reactTags;
-      const trueInput = reactTags.input.input;
-      const forceKey = reactTags.handleKeyDown.bind(reactTags);
-
-      trueInput.addEventListener("keyup", evt => {
-        // did we type a comma?
-        const key = evt.key;
-        if (key === ',') {
-          // remove the comma and type an enter instead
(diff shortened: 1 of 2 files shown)


pages/add/form/detail-info-fields.js
```diff
@@ -80,15 +51,17 @@ let Tags = React.createClass({
     }).filter(suggestion => !!suggestion);
   },
   render: function() {
+    let delimiters = [9,13,32,188]; // keycodes for tab,enter,space,comma
```
Pomax: these are constants, so it's better if, at the top the file, we say:
```javascript
const delimiters = ['\t', '\r', '\n', ' ', ','].map( v => v.charCodeAt(0) );
```

mmmavis: hey @Pomax , I found out the value returned from `charCodeAt` isn’t necessarily the same as the `keycode` returned from keyboard event. Comma is one example - it is `44` from `charCodeAt` but its keycode is `188`.  I searched it a bit but couldn't find a solution... do you know a way to convert character to keycode? If not I guess what I can do is to move that line from `render()` to the top of the file and make it a `const`.

Pomax: Interesting point, I had a look and `keyCode` is apparently obsolete. I've filed this over on https://github.com/i-like-robots/react-tags/issues/81 too.

To resolve this rather than `charCodeAt(0)` we can use `codePointAt(0)` which yields the correct value.

mmmavis: @Pomax `",".codePointAt(0)` returns `44` as well 👀

Pomax: Yeah okay this is weird. Let's take out "32" for space because spaces in tags should be fine (e.g. "digital literacy"), and move it up as a `const DELIMITERD = [...]` and then it's R+

mmmavis:
🍻

PR #590, opened 2017-06-23
Added issues descriptions to /issues plus sorted issues alphabetically

### LINKED ISSUE #516: Implement add issue descriptions to issues tab
Based on design in https://github.com/mozilla/network-pulse/issues/204

This design adds content to what is currently on https://www.mozillapulse.org/issues

Copy is pulled from https://www.mozilla.org/en-US/foundation/issues/ and @alanmoo says he already has the description for each issue in the database, so you don't have to manually paste them in to a route.

Click on the _issue tag_ or the _issue title_ should show the filtered view e.g. https://www.mozillapulse.org/issues/online-privacy-and-security

desktop - issues landing 2x

mmmavis:
Fixes #516

@xmatthewx @kristinashu I don't think it'll matter much but just wanna keep you both in loop - I decided to sort the issues alphabetically so they are not in random order.

screen shot 2017-06-23 at 16 16 06-fullpage

For the sections... instead of listing issues vertically into 2 columns
```
1   4
2   5
3
```
I arranged them horizontally
```
1   2
3   4
5
```

Personally I feel that arrangement is more natural and it works better for smaller screen. 

Let me know if anything needs adjustment! 😉


mmmavis:
@Pomax a note - before we merge this PR we should make sure https://github.com/mozilla/network-pulse-api/pull/140 is live on production!

components/issue-selector/issue-selector.jsx
```diff
@@ -1,15 +1,27 @@
 import React from 'react';
 import NavLink from '../nav-link/nav-link.jsx';
+import Utility from '../../js/utility';
+
+const ISSUES = [
```
Pomax: Given that we get the list of known issues from the `/issues` REST endpoint, including the description, does it make more sense to no longer hardcode these labels and instead add these labels the API server for each issue? (we can do that in follow-up though)

pages/issues/issues.jsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  componentDidMount() {
+    this.getIssues();
+  }
+
+  getIssues() {
+    Service.issues
+      .get()
+      .then((issues) => {
```
Pomax: no need for `(...)` around single arguments.

pages/issues/issues.jsx
```diff
@@ -0,0 +1,61 @@
...
+        // sort issues alphabetically by their name
+        issues.sort((a,b) => {
```
Pomax: no need for the function, `array.sort()` is formally declared as a string sort, so just `issues.sort()` is actually what we want here.

pages/issues/issues.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+  render() {
+    return (
+      <div className="issues-page">
+        <Helmet><title>Issues</title></Helmet>
+        <IssueSelector />
+        <p className="intro">Mozilla has identified five issues that we believe must be tackled in the current era, to build the open internet we want. Tap above to browse by issue.</p>
+        <div className="row">
+          {
+            this.state.issues.map(issue => { return this.renderIssueSection(issue); })
```
Pomax: single line arrow functions autoreturn, so you actually want to use the following syntax:
```
this.state.issues.map(issue => this.renderIssueSection(issue))
```

pages/issues/issues.jsx
mmmavis: Ahhhh I now recall why I didn't go with `issues.sort()`. `issues` should be an array of issue object `{name: ``, description: ``}` and not just an array of issue name strings.

components/issue-selector/issue-selector.jsx
mmmavis: let's tackle it in another ticket: https://github.com/mozilla/network-pulse/issues/596

Pomax:
@mmmavis this looks to work well (I had to updated network-pulse-api because it wasn't correctly serving the name+desc for issues. That's been fixed), but should dedicated pages also show that description?

Right now `/issues` will show name + desc, but `issues/decentralization` does not have a little heading with the description for `decentralization`.

mmmavis:
@Pomax 

> Right now `/issues` will show name + desc, but `issues/decentralization` does not have a little heading with the description for `decentralization`.

Do you mind creating a ticket for this? There might be some ongoing (re)design work that we are not aware of. This would be a quick followup PR anyway but I would like to loop Matthew and Kristina in before we implement this.

Pomax:
Filed #597 as followup

mmmavis:
d(`･∀･)b

PR #593, opened 2017-07-05
disable image link

### LINKED ISSUE #519: Disable image link on detail view
Steps to repeat:
- go to a post detail page e.g. https://www.mozillapulse.org/entry/251
- click on the image and the page will reload

Expected result:
- clicking on image should not reload the page

Sudip7:
Fixes #519

### DIFF
--- a/components/navbar/navbar.jsx
+++ b/components/navbar/navbar.jsx
@@ -23,7 +23,7 @@ class NavBar extends React.Component {
         <div className="container">
           <div className="row">
             <div className="logo col-md-4 mb-3 mb-sm-0">
-              <IndexLink to="/" className="d-inline-block"><img src="/assets/svg/pulse-wordmark.svg" width="204" height="34" alt="Mozilla Pulse" /></IndexLink>
+              <li className="d-inline-block"><img src="/assets/svg/pulse-wordmark.svg" width="204" height="34" alt="Mozilla Pulse" /></li>
             </div>
             <ul className="nav-link-list col-md-8 mb-0">
               <NavListItem><NavLink to="/featured" className="text-nav-link">Featured</NavLink></NavListItem>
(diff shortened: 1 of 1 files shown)


components/navbar/navbar.jsx
```diff
@@ -23,7 +23,7 @@ class NavBar extends React.Component {
         <div className="container">
           <div className="row">
             <div className="logo col-md-4 mb-3 mb-sm-0">
-              <IndexLink to="/" className="d-inline-block"><img src="/assets/svg/pulse-wordmark.svg" width="204" height="34" alt="Mozilla Pulse" /></IndexLink>
+              <li className="d-inline-block"><img src="/assets/svg/pulse-wordmark.svg" width="204" height="34" alt="Mozilla Pulse" /></li>
```
Pomax: `<li>` are list items, they only go inside lists (ordered, `<ol>`, or unordered, `<ul>`). In this case we have `<div><IndexLink></IndexLink></div>` so just remove the surrounding `<IndexLink>`, or if we need that `d-inline-block` class applied, we can see if we can apply it directly to the parent div, or to the img. If both of those look wrong, then instead of removing it entirely, turn it into a `<div className="d-inline-block">  </div>`

Pomax:
Reading through https://github.com/mozilla/network-pulse/issues/519 this does not address the issue that is being talked about. The issue is not about our logo, it's about pulse entries.

STR:
- visit https://www.mozillapulse.org/entry/20
- click on the entry's image, which links to the page you are on right now
- the page reloads

When viewing an entry on its own, rather than as part of a list, the image should not link to its own entry (but it **should** link to its own entry as part of a listing).

Sudip7:
@Pomax : I believe, I misunderstood the requirement. Is it talking about the image in the entry not the "mozilla" logo. 

But I could not completely understand your statement : "When viewing an entry on its own, rather than as part of a list, the image should not link to its own entry (but it should link to its own entry as part of a listing).:"

Pomax:
It turns out that the work in https://github.com/mozilla/network-pulse/pull/575 (which is a big PR and will be landing today) also fixed this issue, so this PR is no longer necessary.

PR #610, opened 2017-07-20
Bookmark entries with api

### LINKED ISSUE #310: Use Pulse API to store entries' fav state
Currently Pulse uses `localStorage` to store a list of IDs of the entries that a user has fav'ed. Once the API is ready we should use the API to store this info on the database instead.

Question - Do we want to show heart(fav) icon ❤️ before users are logged in? If so, what does the icon do? Does clicking on it bring users to the login page? or?

/cc @kristinashu @Pomax @alanmoo

mmmavis:
Fixes #310 

❗️❗️❗️ Dependent on https://github.com/mozilla/network-pulse-api/pull/156 . Don't merge this PR until we make sure https://github.com/mozilla/network-pulse-api/pull/156 is pushed on API production.

---

@Pomax there are lots of changes happening in `bookmarks.jsx` and `project-card.jsx`. It might be easier to review these 2 files by looking at the files directly instead of checking the code differences that were introduced in this PR.

### DIFF
--- a/components/project-card/project-card.jsx
+++ b/components/project-card/project-card.jsx
@@ -83,38 +83,16 @@ class ProjectCard extends React.Component {
     }
   }
 
-  getNonce(callback) {
-    Service.nonce()
-      .then((nonce) => {
-        callback(false, nonce);
-      })
-      .catch((reason) => {
-        callback(new Error(`Could not retrieve data from /nonce. Reason: ${reason}`));
-      });
-  }
-
   toggleBookmark(callback) {
-    this.getNonce((error, nonce) => {
-      if (error) {
-        console.error(error);
-        return;
-      }
-
-      let formattedNonce = {
-        nonce: nonce.nonce,
-        csrfmiddlewaretoken: nonce.csrf_token
-      };
-
-      Service.entry
-        .put.bookmark(this.props.id, formattedNonce)
-        .then(() => {
-          callback(null);
-        })
-        .catch(reason => {
-          console.error(reason);
-          callback(reason);
-        });
-    });
+    Service.entry
+      .put.bookmark(this.props.id)
+      .then(() => {
+        callback(null);
+      })
+      .catch(reason => {
+        console.error(reason);
+        callback(reason);
+      });
   }
 
   handleBookmarkClick() {
--- a/components/project-card/project-card.jsx
+++ b/components/project-card/project-card.jsx
@@ -60,24 +60,24 @@ class ProjectCard extends React.Component {
     }
   }
 
-  bookmarkProjectOnLs(bookmarks) {
+  bookmarkToLocalStorage(bookmarks) {
     bookmarks.unshift(this.props.id);
     this.setState({bookmarked: true});
   }
 
-  unbookmarkProjectOnLs(bookmarks) {
+  unbookmarkToLocalStorage(bookmarks) {
     bookmarks.splice(bookmarks.indexOf(this.props.id), 1);
     this.setState({bookmarked: false});
   }
 
-  toggleBookmarkedStateOnLs() {
+  updateBookmarkOnLocalStorage() {
     let bookmarks = bookmarkManager.bookmarks.get();
 
     if (bookmarks) {
       if (this.state.bookmarked) {
-        this.unbookmarkProjectOnLs(bookmarks);
+        this.unbookmarkToLocalStorage(bookmarks);
       } else {
-        this.bookmarkProjectOnLs(bookmarks);
+        this.bookmarkToLocalStorage(bookmarks);
       }
       bookmarkManager.bookmarks.set(bookmarks);
     }
@@ -93,7 +93,7 @@ class ProjectCard extends React.Component {
       });
   }
 
-  toggleBookmarkedStateOnDb(callback) {
+  toggleBookmark(callback) {
     this.getNonce((error, nonce) => {
       if (error) {
         console.error(error);
@@ -117,7 +117,7 @@ class ProjectCard extends React.Component {
     });
   }
 
-  toggleBookmarkedState() {
+  handleBookmarkClick() {
     if (document && document.onanimationend !== `undefined`) {
       this.refs.heart.classList.add(`beating`);
       this.refs.heart.addEventListener(`animationend`, () => {
@@ -128,7 +128,7 @@ class ProjectCard extends React.Component {
     ReactGA.event(this.createGaEventConfig(`Bookmark button`, this.state.bookmarked ? `Unbookmarked` : `Bookmarked`));
 
     if (user.loggedin) {
-      this.toggleBookmarkedStateOnDb((error) => {
+      this.toggleBookmark((error) => {
         if (error) return;
 
         this.setState({
@@ -139,7 +139,7 @@ class ProjectCard extends React.Component {
       return;
     }
 
-    this.toggleBookmarkedStateOnLs();
+    this.updateBookmarkOnLocalStorage();
   }
 
   handleThumbnailClick() {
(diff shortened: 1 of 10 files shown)


pages/bookmarks.jsx
```diff
@@ -45,50 +48,113 @@ export default React.createClass({
...
     }
 
-    return bookmarkedProjects;
+    return <div>
+            {signInPrompt}
+            {bookmarkedProjects}
+          </div>;
+  },
+  renderContent() {
+    if (user.loggedin === undefined) return null;
```
mmmavis: It takes a few seconds to find out if the current user is logged in or not. We shouldn't assume the current user is anonymous and show him/her "Please sign in" message. I changed the default value for `user.loggedin` to `null` (instead of `false`) so we can show a blank screen while getting result back from the `/userstatus` call.

js/app-user.js
```diff
@@ -80,9 +80,9 @@ class User {
     this.resetUser();
   }
 
-  resetUser() {
+  resetUser(justLoggedOut) {
     this.username = undefined;
```
mmmavis: It takes a few seconds to find out if the current user is logged in or not. We shouldn't assume the current user is anonymous by setting `user.loggedin` to `false`.

js/bookmarks-manager.js
```diff
@@ -1,22 +1,38 @@
-export const getBookmarks = function() {
```
mmmavis: Same code just with some cleanup.

pages/add/add.jsx
```diff
@@ -190,32 +190,30 @@ export default React.createClass({
...
                          linkComponent={linkComponent}>
-            </HintMessage>);
+              {additionalMessage}
+            </HintMessage>;
   },
   getContent() {
-    if (user.loggedin) {
-      return this.getContentForLoggedInUser();
-    }
+    if (user.loggedin === undefined) return null;
```
mmmavis: It takes a few seconds to find out if the current user is logged in or not. We shouldn't assume the current user is anonymous and show him/her "Please sign in" message. I changed the default value for `user.loggedin` to `null` (instead of `false`) so we can show a blank screen while getting result back from the `/userstatus` call.

Pomax:
An interesting edge case I run into is one where my localStorage contains entry ids that don't exist (from older tests). In that case, localStorage contains:
```
1,2,4,8,10
```
and the database only contains
```
1,2
```
so when it tries to sync, it will set the bookmarks for 1 and 2, and then throw a 400 error when it sees 4, which is a non-existent entry.

components/project-card/project-card.jsx
```diff
@@ -41,33 +43,80 @@ class ProjectCard extends React.Component {
...
-        bookmarked = false;
-      }
+      bookmarked = bookmarks.indexOf(this.props.id) > -1;
       this.setState({bookmarked: bookmarked});
     }
   }
 
-  bookmarkProject(bookmarks) {
-    ReactGA.event(this.createGaEventConfig(`Bookmark button`, `Bookmarked`));
+  bookmarkProjectOnLs(bookmarks) {
```
Pomax: Minification means function names don't matter. It is *much* better to call this `bookmarkToLocalStorage` than using "OnLs". New/future devs will have no idea what LS refers to.

components/project-card/project-card.jsx
```diff
@@ -41,33 +43,80 @@ class ProjectCard extends React.Component {
...
 
-  unbookmarkProject(bookmarks,index) {
-    ReactGA.event(this.createGaEventConfig(`Bookmark button`, `Unbookmarked`));
-    bookmarks.splice(index,1);
+  unbookmarkProjectOnLs(bookmarks) {
+    bookmarks.splice(bookmarks.indexOf(this.props.id), 1);
     this.setState({bookmarked: false});
   }
 
+  toggleBookmarkedStateOnLs() {
```
Pomax: this is a bad function name, too, let's give that a more immediately obvious name.

components/project-card/project-card.jsx
```diff
@@ -76,18 +125,21 @@ class ProjectCard extends React.Component {
...
-    if (bookmarks) {
-      index = bookmarks.indexOf(this.props.id);
-      if (index > -1) {
-        this.unbookmarkProject(bookmarks,index);
-      } else {
-        this.bookmarkProject(bookmarks);
-      }
-      saveBookmarks(bookmarks);
+    if (user.loggedin) {
+      this.toggleBookmarkedStateOnDb((error) => {
```
Pomax: the client doesn't know anything about databases, it just knows about a server. That said, it is better to assume that "without any qualifiers", we're talking about the server so that `toggleBookmark` goes to the server, and `toggleBookmarkInLocalStorage` doesn't.

components/project-card/project-card.jsx
```diff
@@ -41,33 +43,80 @@ class ProjectCard extends React.Component {
...
+      if (this.state.bookmarked) {
+        this.unbookmarkProjectOnLs(bookmarks);
+      } else {
+        this.bookmarkProjectOnLs(bookmarks);
+      }
+      bookmarkManager.bookmarks.set(bookmarks);
+    }
+  }
+
+  getNonce(callback) {
```
Pomax: don't we have code for this elsewhere already? If so, are we repeating code here?

pages/bookmarks.jsx
```diff
@@ -45,50 +48,92 @@ export default React.createClass({
...
+        onError(null);
+      })
+      .catch(reason => {
+        onError(reason);
+      });
   },
-  getAnonymousContent() {
-    return (<p><button className="btn btn-link inline-link" onClick={this.handleSignInBtnClick}>Sign in with Google</button>.</p>);
+  handleImportBookmarksClick() {
+    console.log(`handleImportBookmarksClick`);
```
Pomax: remove?

pages/bookmarks.jsx
```diff
@@ -45,50 +48,92 @@ export default React.createClass({
...
+      }
+    });
+
+    // and clear ls bookmarks
   },
-  getloggedInStatusContent() {
-    if (user.loggedin) {
-      return this.getContentForLoggedInUser();
+  getContentForLoggedInUser() {
+    // console.log(`> getContentForLoggedInUser`);
```
Pomax: better to remove rather than keep commented

pages/bookmarks.jsx
```diff
@@ -45,50 +48,92 @@ export default React.createClass({
...
-    // For user facing bits like UI labels and URL path we want them to stay as "favorites".
-    // For more info see: https://github.com/mozilla/network-pulse/issues/326
-    let headerText = `Save your Favs`;
-    let bookmarkedProjects = (<HintMessage iconComponent={<img src="/assets/svg/icon-bookmark-selected.svg" />}
-                                           header={headerText}
-                                           linkComponent={<Link to={`/featured`}>Explore featured</Link>}>
-                                <p>Tap the heart on any project to save it here.</p>
-                              </HintMessage>);
+  getAnonymousContent() {
+    // console.log(`> getAnonymousContent`);
```
Pomax: remove comment?

js/bookmarks-manager.js
```diff
@@ -1,22 +1,35 @@
...
+    set: function(bookmarksArray) {
+      try {
+        let bookmarks = JSON.stringify(bookmarksArray);
 
-    localStorage.setItem(`bookmarks`, bookmarks );
-  } catch (err) {
-    console.log(err);
+        localStorage.setItem(LS_BOOKMARKS_KEY, bookmarks);
+      } catch (err) {
+        console.log(err);
```
Pomax: should this be `console.error(err)` instead?

js/bookmarks-manager.js
```diff
@@ -1,22 +1,35 @@
...
+      }
+    },
+    delete: function() {
+      try {
+        localStorage.removeItem(LS_BOOKMARKS_KEY);
+      } catch (err) {
+        console.log(err);
```
Pomax: should this be `console.error(err)` instead?

mmmavis:
🎉 🎉 🎉 🎉 🎉

Issue #628, opened 2017-08-01
Update dark grey to black


kristinashu:
[New Mozilla brand](https://mozilla.ninja/) uses black instead of dark grey. We need to update the 'add project' icon and the footer to black.

https://redpen.io/projects

Related to font update https://github.com/mozilla/network-pulse/issues/614

Sudip7:
hey @kristinashu  though I don't have much knowledge on design, I would like take this ticket. Would you please assist in location the css files for this project. I was looking at https://github.com/mmmavis/network-pulse/tree/gh-pages/css to get some idea but similar structure is not available here. Please assist.

kristinashu:
Awesome, it would be great if you did this @Sudip7! Unfortunately I don't know where the file is either. @mmmavis are you able to point it out? Is it related to our [Mofo Bootstrap](https://github.com/mozilla/mofo-bootstrap)?

Sudip7:
@kristinashu , m not sure but I believe it's probably a  forked repository or a new one created by her. 
@mmmavis  would you please assist me on this. Then I may able to try the #614 .

mmmavis:
For changing background colour of the footer we should update [mofo-bootstrap](https://github.com/mozilla/mofo-bootstrap) so this colour change will reflect on other sites like the Network site. (I'll create a ticket on mofo-bootstrap and self assign)

For changing the colour of the "Add Project" button on Network Pulse it'll only require updating a [colour value](https://github.com/mozilla/network-pulse/blob/93492bdb0d5cb5633be24e1c4281a1f3036622ae/components/navbar/navbar.scss#L51) in a `SCSS` file.

@Sudip7 the `gh-pages` branch on my forked repo was for Pulse v1 (jQuery-based, uses vanilla CSS). We are no longer updating that branch and the branch you should be looking at would be the `master` branch of this Mozilla repo 😃 Pulse is now React-based and uses SCSS which its syntax is pretty CSS-alike. Let me know if you'd like to work on updating the colour of the "Add Project" button (the plus sign button at top right of Pulse page)!

Sudip7:
I would love to do that. Should I look for main.scss or any other file to change the font or color. Pulse v1 repo css config was bit straight forward but still I would like to dirty my hands. Please assist.

mmmavis:
Hey @Sudip7 take a look at my previous comment. I've included a link that brings you directly to the line of code that needs changing. This one should be straightforward to fix.

Sudip7:
[...]

PR #637, opened 2017-08-05
changing add btn color to black

Sudip7:
#628 : add btn color changed to black


mmmavis:
@Sudip7 code looks good! Thanks!

@kristinashu are we okay with changing the add button colour to black and keeping the footer background unchanged?

kristinashu:
Yup, fine with me!

Issue #622, opened 2017-08-01
Build MVP profile view in client


alanmoo:

xmatthewx:
MVP is person's details and simple list of all associated entries below. 

https://redpen.io/qua34b5932cdb6cf9f

kristinashu:
I believe this is what the MVP profile page would look like and should include:

- edit profile which links to [form](https://redpen.io/bk7c2a3cbaafdcd47a)
- all projects the person/organization has uploaded and/or is listed as a creator on (no duplicates)

Info the user adds:
- profile pic
- name (from google account but updatable in user form) - this is the only required
- location (city, country) - not sure if MVP is clickable?
- languages (multiple separated my commas)
- website URL
- bio (140 character limit)
- clickable 'interest' tags
- clickable 'can help with' tags

Admin adds via django:
- role 
- URL to story engine interview

![full profile mvp 2x](https://user-images.githubusercontent.com/1068783/28890635-893418d4-777c-11e7-9d5b-62994875bf44.png)

For reference, here is the form:
![profile form 2x](https://user-images.githubusercontent.com/1068783/28890365-b02219f6-777b-11e7-82a6-92e644fd447c.png)

mmmavis:
@kristinashu thanks for the list. Could you provide mockup for mobile view? I'm afraid the following section will look super clustered if you are letting me design it haha.

<img width="553" alt="screen shot 2017-08-02 at 12 41 33 pm" src="https://user-images.githubusercontent.com/2896608/28891612-f652dd3a-777f-11e7-86f9-423054522841.png">

kristinashu:
How about something like this?

<img width="316" alt="screen shot 2017-08-03 at 8 24 00 am" src="https://user-images.githubusercontent.com/1068783/28929568-23873756-7825-11e7-8e27-c03f1ed7ca87.png">

mmmavis:
Thanks @kristinashu 

I wonder if we should break `language`, `Read Story`, `link` into its own line with an icon like what you have for the `location` field. If an user has indicated more than 1 language or the full url of the link is long, these meta mostly likely won't fit into one line.

kristinashu:
Yeah, that sounds good! I'll work on the icons.

(note that most people won't actually fill in all that info)

mmmavis:
> Yeah, that sounds good! I'll work on the icons.

We can use font-awesome icons if you'd like.

kristinashu:
[...]

PR #639, opened 2017-08-08
Profile Page - render MVP view for an user's own profile

mmmavis:
Related to #622

**This PR is a feature branch against another feature branch `all-things-profile-related`.**

It only covers code needed to render a loggedin user's own profile view

- visit the page by clicking the "Hi, [username]" link on nav bar
- or visiting `/profile/me` directly

Everything profile related features/specs will be tackled in follow-up PRs.

### DIFF
--- a/components/bio/bio.jsx
+++ b/components/bio/bio.jsx
@@ -16,7 +16,7 @@ class Bio extends React.Component {
   renderEditLink() {
     if (!this.props.my_profile) return null;
 
-    return <div className="mt-3"><Link to="/">edit profile</Link></div>;
+    return <div className="mt-3"><Link to="/myprofile">edit profile</Link></div>;
   }
 
   renderName() {
--- a/components/bio/bio.jsx
+++ b/components/bio/bio.jsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import { Helmet } from "react-helmet";
+// import { Link } from 'react-router';
+
+class Bio extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  renderPic() {
+    let style = {
+      backgroundImage: `url(${this.props.pic})`,
+    };
+
+    return <div className="pic mx-4" style={style}></div>;
+  }
+
+  renderName() {
+    return <div className="name d-inline-block mr-4">{this.props.firstname} {this.props.lastname}</div>;
+  }
+
+  renderSocialMedia() {
+    if (!this.props.socialMedia || this.props.socialMedia.length < 1) return null;
+
+    let platforms = this.props.socialMedia.map(platform => {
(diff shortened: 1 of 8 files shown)


components/bio/bio.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+    return this.renderTags(`interests`, `issues`, this.props.issues);
+  }
+
+  render() {
+    return (
+      <div className="bio mb-5">
+        <div className="row">
+          <div className="col-sm-4 col-md-2 text-center">
+            { this.renderThumbnail() }
```
gvn: Usually I put functions to render JSX within the render method instead of as separate methods, but I guess it doesn't really matter.

components/bio/bio.jsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  renderThumbnail() {
+    return <div className="thumbnail mx-auto">
+            { this.props.thumbnail && <img src={this.props.thumbnail} className="img-fluid" /> }
+          </div>;
+  }
+
+  renderEditLink() {
+    if (!this.props.my_profile) return null;
```
gvn: Could do this whole function body as a single ternary operator

components/bio/bio.jsx
```diff
@@ -0,0 +1,124 @@
...
+            { this.renderBlurb() }
+            { this.renderInterest() }
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+Bio.propTypes = {
```
gvn: Why isn't this declared inside the class?

components/navbar/navbar.jsx
```diff
@@ -39,28 +39,35 @@ class NavBar extends React.Component {
     return <NavListItem><NavLink to="/moderation">Moderation</NavLink></NavListItem>;
   }
 
+  renderUsername() {
+    if (!user.loggedin) return null;
+    console.log(user);
```
gvn: Remove console.log

components/navbar/navbar.jsx
```diff
@@ -39,28 +39,35 @@ class NavBar extends React.Component {
     return <NavListItem><NavLink to="/moderation">Moderation</NavLink></NavListItem>;
   }
 
+  renderUsername() {
+    if (!user.loggedin) return null;
```
gvn: Another place a ternary would work

components/navbar/navbar.jsx
mmmavis: :D

components/bio/bio.jsx
mmmavis: The "old way" is going to be deprecated soon. 

> React.PropTypes has moved into a different package since React v15.5. Please use the prop-types library instead

PR #640, opened 2017-08-09
Added featured toggle to Moderation page

### LINKED ISSUE #616: Implement moderator Feature toggle
Design moderator Feature toggle #602 
API endpoint in #572

On https://www.mozillapulse.org/moderation with "Approved" dropdown selected, a users will see an additional check box. If they check it then that post will be featured.
screen shot 2017-07-26 at 9 56 45 am

We also need to add "Featured" to this v drop down so users only see what is featured there (and can uncheck posts).

screen shot 2017-07-26 at 9 56 02 am

mmmavis:
Fixes #616 

❗️Before merging, make sure the new env var `LEARN_MORE_LINK` is set

### DIFF
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -36,14 +36,14 @@ class Search extends React.Component {
     document.querySelector(`#search-box`).focus();
   }
 
-  getSearchCriteria(theProps) {
+  getSearchCriteria(props) {
     let featured = FEATURE_FILTER_OPTIONS.filter(option => {
-      return theProps.location.query.featured === option.value;
+      return props.location.query.featured === option.value;
     })[0];
 
     return {
-      keywordSearched: theProps.location.query.keyword,
-      moderationState: { value: ``, label: theProps.location.query.moderationstate || DEFALUT_MODERATION_FILTER },
+      keywordSearched: props.location.query.keyword,
+      moderationState: { value: ``, label: props.location.query.moderationstate || DEFALUT_MODERATION_FILTER },
       featured: featured || { value: ``, label: DEFALUT_FEATURED_FILTER }
     };
   }
@@ -111,7 +111,8 @@ class Search extends React.Component {
                               type="search"
                               onChange={(event) => this.handleInputChange(event)}
                               placeholder="Search keywords, people, tags..."
-                              className="form-control" />
+                              className="form-control"
+              />
               <button className="btn dismiss" onClick={() => this.handleDismissBtnClick()}>&times;</button>
             </div>
           </div>;
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -9,6 +9,12 @@ import Service from '../../js/service.js';
 import ProjectLoader from '../../components/project-loader/project-loader.jsx';
 
 const DEFALUT_MODERATION_FILTER = `Pending`;
+const DEFALUT_FEATURED_FILTER = `---`;
+const FEATURE_FILTER_OPTIONS = [
+  { value: `True`, label: `Featured` },
+  { value: `False`, label: `Not featured` },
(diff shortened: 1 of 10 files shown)


pages/search/search.jsx
```diff
@@ -117,36 +130,52 @@ class Search extends React.Component {
...
-                className="state-filter d-inline-block text-left"
-                searchable={false}
-                clearable={false}
-                cache={false}
-                placeholder="Moderation state"
-                loadOptions={(input, callback) => this.getModerationStates(input, callback)}
-                onChange={(selected) => this.handleFilterChange(selected)}
-              />
-            </div>;
+    return <Select.Async
```
mmmavis: removed wrapper `div`

mmmavis:
@Pomax I messed up `rebase` again 😭 . Will ping you once I fix those issues. Please don't review this PR until then.

mmmavis:
@Pomax this is good for review now! 🍻

pages/search/search.jsx
```diff
@@ -9,6 +9,12 @@ import Service from '../../js/service.js';
 import ProjectLoader from '../../components/project-loader/project-loader.jsx';
 
 const DEFALUT_MODERATION_FILTER = `Pending`;
+const DEFALUT_FEATURED_FILTER = `---`;
```
Pomax: typo in "DEFALUT". But also this does not seem the right way to offer this filter. Searching for things inside featured makes sense, but searching for things "only in not-featured" seems really weird. It would make more sense to have a checkbox or button toggle here that either restricts search to featured-only, or doesn't do any filtering based on feature status at all, simply hitting all entries irrespective of feature flag.

pages/search/search.jsx
```diff
@@ -31,15 +37,21 @@ class Search extends React.Component {
   }
 
   getSearchCriteria(theProps) {
+    let featured = FEATURE_FILTER_OPTIONS.filter(option => {
+      return theProps.location.query.featured === option.value;
```
Pomax: can we rename `theProps` to just `props`? "the" as an article from natural English is already implied by virtue of this being a defined object being passed as input argument.

pages/search/search.jsx
```diff
@@ -81,22 +99,19 @@ class Search extends React.Component {
...
-            {label}
             <div className={classnames}>
               <DebounceInput id="search-box"
                               value={this.state.keywordSearched}
                               debounceTimeout={300}
                               type="search"
                               onChange={(event) => this.handleInputChange(event)}
-                              placeholder="Search keywords, people, tags..." />
+                              placeholder="Search keywords, people, tags..."
+                              className="form-control" />
```
Pomax: worth treating the `/>` the same as you would a `}`, so that you don't have to move around the `/>` when you add/remove properties:
```
<DebounceInput
  id="search-box"
  value={this.state.keywordSearched}
  debounceTimeout={300}
  type="search"
  onChange={(event) => this.handleInputChange(event)}
  placeholder="Search keywords, people, tags..."
  className="form-control"
/>
```

routes.jsx
```diff
@@ -21,6 +21,7 @@ import Footer from './components/footer/footer.jsx';
 const Featured = () => {
   return <div>
           <Helmet><title>Featured</title></Helmet>
+          <p>Discover & collaborate on projects for a healthy internet. <a href="https://www.mozillapulse.org/entry/120">Learn more</a>.</p>
```
Pomax: Can we turn this into an env var instead of hardcoding the entry URL? If we ever have to do database maintenance there is no guarantee that the ids stay the same.

pages/search/search.jsx
mmmavis: Yea I was thinking if we should give moderators more options to filter but now I think about it the not-featured option might not be used much.

Pomax:
looks good to me now

Issue #485, opened 2017-04-13
Detail view corrections link


xmatthewx:
Since some things will be posted on behalf of others, would be good to give people the means to correct stuff. We have contact in the footer. This is a bit more accessible. 

<img width="687" alt="screen shot 2017-04-13 at 6 03 49 pm" src="https://cloud.githubusercontent.com/assets/53411/25026272/c1e669e4-2073-11e7-91b3-cec0aa5afb34.png">

Should be unobtrusive. This comp uses half opacity.

kristinashu:
This looks good. What would clicking Contact Us do (email us, a form, open issue...)? 

This area is also where people could flag a post.

xmatthewx:
Yes. We have a contact link in the footer that goes to a google form.

mmmavis:
Do we still wanna add this? @kristinashu @xmatthewx this should be a quick fix.

kristinashu:
I think so!

(slightly related to Enable users to edit their own entries #472)

xmatthewx:
Yes. We'll have to think through the form, but could just start with the generic contact form in the footer.

mmmavis:
> Yes. We'll have to think through the form, but could just start with the generic contact form in the footer.

Cool, I'll have a PR for this today.

mmmavis:
Question is where to add the link on the new 2-column layout? https://network-pulse-staging.herokuapp.com/entry/334

kristinashu:
Good questions... how about we add a diving line then have the text just under it?

![screencapture-network-pulse-staging-herokuapp-entry-334-1502313850315](https://user-images.githubusercontent.com/1068783/29144759-fe29a6da-7d0e-11e7-93c5-3027272c33ef.jpg)

mmmavis:
Live on production!

PR #642, opened 2017-08-09
Added correction link

mmmavis:
Related to #485 

@gvn The report correct link will only show on detail view (e.g., http://localhost:3000/entry/334) and not list view (e.g., http://localhost:3000/latest )


mmmavis:
Yeah! Thanks @gvn

Issue #645, opened 2017-08-11
Refine Get Involved for empty fields 


xmatthewx:
If an entry has no Get Involved elements (text, link, categories), we shouldn't display the Get Involved header. 

The first two entries submitted that included Get Involved categories, didn't have text or a link. The help tags almost look like buttons to Get involved. I wonder if there's  placeholder text we could put in there in this scenario with categories but no text, no link. 

<img width="831" alt="screen shot 2017-08-11 at 2 31 18 pm" src="https://user-images.githubusercontent.com/53411/29227180-dc63e89a-7ea2-11e7-82e8-b7045ae43ebf.png">

xmatthewx:
Scenarios
 1. no link, text, categories => hide section
 2. categories only => fallback text?
 3. text or link only => display normal?

### 2 fallback text

does this work?

<img width="832" alt="screen shot 2017-08-11 at 2 38 29 pm" src="https://user-images.githubusercontent.com/53411/29227311-67a1019a-7ea3-11e7-8330-05ea20ea0739.png">

kristinashu:
Seems like a good solution to me!

xmatthewx:
@mmmavis - we have a bit of follow up on this to deal with entries with missing "get involved" info. 

Scenario => What we display  
 - no link, no text, no categories => hide section
 - text or link only => display normal
 - categories only => display fallback text: "Help with this, or find other projects that have similar ways to get involved." (see mockup above)

kristinashu:
@xmatthewx can you please also review this one? Everything is much better and showing up as you suggested. The only case where it still feels slightly odd is when there is only a link. So we see the title "Get involved" followed by link "Get involved" e.g. https://network-pulse-staging.herokuapp.com/entry/201 Is that ok?

mmmavis:
https://network-pulse-staging.herokuapp.com/entry/336
 https://network-pulse-staging.herokuapp.com/entry/336
 https://network-pulse-staging.herokuapp.com/entry/336
 https://network-pulse-staging.herokuapp.com/entry/336

mmmavis:
I've created some test entries

 https://network-pulse-staging.herokuapp.com/entry/336
 https://network-pulse-staging.herokuapp.com/entry/337
 https://network-pulse-staging.herokuapp.com/entry/338
 https://network-pulse-staging.herokuapp.com/entry/339
 https://network-pulse-staging.herokuapp.com/entry/340

xmatthewx:
[...]

PR #660, opened 2017-08-16
Tidy up get involved section

mmmavis:
Relates to #645 

I have created a few entries on staging for testing (id `336` to `340`)


mmmavis:
🎉

PR #665, opened 2017-08-17
Don't double encode issue name

### LINKED ISSUE #664: Bug: API params got encoded twice
A few issue pages now have problem displaying the right set of data. This is because the params got encoded twice

e.g.,
- https://api.mozillapulse.org/entries/?issue=Digital%2520Inclusion
- should really be https://api.mozillapulse.org/entries/?issue=Digital%20Inclusion

I will submit a patch soon.

mmmavis:
Fixes #664

### DIFF
--- a/components/project-loader/project-loader.jsx
+++ b/components/project-loader/project-loader.jsx
@@ -67,6 +67,10 @@ export default React.createClass({
       delete combinedParams.bookmarkedOnly;
     }
 
+    if (combinedParams.showCounter) {
+      // showCounter is not a query param the API supports
+      delete combinedParams.showCounter;
+    }
 
     return Object.assign(combinedParams, { page: this.state.nextBatchIndex });
   },
--- a/components/project-loader/project-loader.jsx
+++ b/components/project-loader/project-loader.jsx
@@ -67,11 +67,6 @@ export default React.createClass({
       delete combinedParams.bookmarkedOnly;
     }
 
-    if (combinedParams.showCounter) {
-      // showCounter is not a query param the API supports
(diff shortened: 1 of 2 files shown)


components/project-loader/project-loader.jsx
```diff
@@ -67,6 +67,10 @@ export default React.createClass({
       delete combinedParams.bookmarkedOnly;
     }
 
+    if (combinedParams.showCounter) {
```
Pomax: do we need to delete though? the API should just ignore unknown args

pages/issue.jsx
```diff
@@ -10,6 +10,6 @@ export default function (props) {
   return <div>
             <Helmet><title>{issueName}</title></Helmet>
             <IssueSelector />
-            <ProjectLoader issue={encodeURIComponent(issueName)} showCounter={true} />
+            <ProjectLoader issue={issueName} showCounter={true} />
```
Pomax: oh man, is there a way for us to add a test to the code that will catch things like this in the future?

components/project-loader/project-loader.jsx
mmmavis: not really but i thought this would make it a cleaner url

pages/issue.jsx
mmmavis: yea, i created a ticket: https://github.com/mozilla/network-pulse/issues/666

components/project-loader/project-loader.jsx
Pomax: usually it's better to not call `delete`, so I'd say let's leave this out.

PR #669, opened 2017-08-17
More ES6ing things

mmmavis:


pages/add/form/get-help-fields.js
```diff
@@ -2,14 +2,16 @@ import React from 'react';
 import validator from './validator';
 import Service from '../../../js/service';
 
-let HelpTypes = React.createClass({
-  getInitialState() {
-    return {
+class HelpTypes extends React.Component {
+  constructor() {
```
Pomax: Important to use `constructor(props) { super(props); ... }` here, if you want to be able to rely on `this.props` in the component at all.

mmmavis: @Pomax I've read some articles online saying if we are not using `props` to set `state` there's no need to pass `props` to `super()`. The `App` class in `routes.jsx` seems to be functioning normally without passing `props` to `super()`.

Pomax: This is a choice. I prefer to acknowledge that the constructor is supposed to take `props`, and let the code figure out what to do based on the fact that it's getting `super(props)` called, concretely, as `super(undefined)` when we don't have a properset in the JSX that creates the object.

You can leave `props` off, but I'd prefer not to.

mmmavis:
🍰 ☕️

PR #670, opened 2017-08-17
Show right set of query params in url

### LINKED ISSUE #643: Search includes moderation param in URL
A search incorrectly appends URL with a moderation param. It does this even if a user is _not_ logged in. Note: this does not effect search results, so it is not broken. 

Steps to Repeat
 - Go to https://www.mozillapulse.org/search
 - Search for "mozfest"
 - URL is now: https://www.mozillapulse.org/search?keyword=mozfest&moderationstate=Pending

If you visit https://www.mozillapulse.org/search?keyword=mozfest directly, parameter does not get added. It only happens with interaction on search form. It's also odd that the param is pending, but pulse displays approved.

mmmavis:
Fixes #643

### DIFF
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -37,8 +37,13 @@ class Search extends React.Component {
 
     if (this.props.moderation) {
       // the following states are only useful on moderation mode
-      if (props.location.query.featured === `True`) criteria.featured = props.location.query.featured;
-      criteria.moderationState = { value: ``, label: props.location.query.moderationstate || DEFAULT_MODERATION_FILTER };
+      if (props.location.query.featured === `True`) {
+        criteria.featured = props.location.query.featured;
+      }
+      criteria.moderationState = {
+        value: ``,
+        label: props.location.query.moderationstate || DEFAULT_MODERATION_FILTER
+      };
     }
 
     return criteria;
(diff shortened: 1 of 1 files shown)


pages/search/search.jsx
```diff
@@ -31,11 +31,17 @@ class Search extends React.Component {
...
-      keywordSearched: props.location.query.keyword,
-      moderationState: { value: ``, label: props.location.query.moderationstate || DEFAULT_MODERATION_FILTER },
-      featured: props.location.query.featured || ``
+    let criteria = {
+      keywordSearched: props.location.query.keyword
     };
+
+    if (this.props.moderation) {
+      // the following states are only useful on moderation mode
+      if (props.location.query.featured === `True`) criteria.featured = props.location.query.featured;
```
Pomax: always better to use `{}` so that it's super obvious what is, and is not, controlled by the `if` statement. Even if it's a one liner

pages/search/search.jsx
```diff
@@ -31,11 +31,17 @@ class Search extends React.Component {
...
+      criteria.moderationState = { value: ``, label: props.location.query.moderationstate || DEFAULT_MODERATION_FILTER };
```
Pomax: Probably better spread out over a few lines.

pages/search/search.jsx
```diff
@@ -45,20 +51,17 @@ class Search extends React.Component {
...
 
-
     if ( keywordSearched ) {
       query.keyword = keywordSearched;
     }
 
-    if ( moderationState ) {
+    if ( this.props.moderation ) {
+      // the following params are only useful on moderation mode
+      if ( featured === `True` ) query.featured = featured;
```
Pomax: same here, adding those `{ ... }` is always a good idea

Pomax:
R+

mmmavis:
wooohoooo 💃

PR #677, opened 2017-08-29
Keyboard should auto dismiss after user hit go/search

### LINKED ISSUE #492: Seach keyboard on mobile should dismiss after hitting search/enter on keyboard
From #419

### Action
On Mobile, hit search/enter key on keyboard

### Expected
Keyboard is dismissed

mmmavis:
Fixes #492 

@alanmoo @gvn I don't have an Android device with me. Can either of you help check? This works fine on iOS 10 latest Safari/FF/Chrome.

To QA,

1. check out this PR
2. `npm start` the app
3. ngrok it so you can visit the app on your mobile
4. go to the search page (by tapping the search icon on the nav)
5. type something in the input box, see if the default "return" key (not sure what the Android equivalent is called) becomes "Go/Search"
6. type in some text and hit that "Go/Search" key, see if keyboard gets auto dismissed after

(note that you won't be seeing any projects unless your ngork instance is whitelisted by the api)

### DIFF
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -104,7 +104,7 @@ class Search extends React.Component {
 
     return <div className="d-flex align-items-center">
             <div className={classnames}>
-              <form action="for-search-key-to-show-on-mobile-keyboard">
+              <form action="for-search-key-to-show-on-mobile-keyboard" role="search">
                 <DebounceInput id="search-box"
                                 value={this.state.keywordSearched}
                                 debounceTimeout={300}
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -104,14 +104,16 @@ class Search extends React.Component {
 
     return <div className="d-flex align-items-center">
             <div className={classnames}>
-              <DebounceInput id="search-box"
-                              value={this.state.keywordSearched}
-                              debounceTimeout={300}
-                              type="search"
-                              onChange={(event) => this.handleInputChange(event)}
-                              placeholder="Search keywords, people, tags..."
-                              className="form-control"
-              />
+              <form action="for-search-key-to-show-on-mobile-keyboard">
+                <DebounceInput id="search-box"
+                                value={this.state.keywordSearched}
+                                debounceTimeout={300}
+                                type="search"
+                                onChange={(event) => this.handleInputChange(event)}
+                                placeholder="Search keywords, people, tags..."
+                                className="form-control"
+                />
+              </form>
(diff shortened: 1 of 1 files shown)


pages/search/search.jsx
```diff
@@ -104,14 +104,16 @@ class Search extends React.Component {
...
             <div className={classnames}>
-              <DebounceInput id="search-box"
-                              value={this.state.keywordSearched}
-                              debounceTimeout={300}
-                              type="search"
-                              onChange={(event) => this.handleInputChange(event)}
-                              placeholder="Search keywords, people, tags..."
-                              className="form-control"
-              />
+              <form action="for-search-key-to-show-on-mobile-keyboard">
```
mmmavis: wrapping `<input type="search" ...>` with `<form action="not-blank">` does the trick

pages/search/search.jsx
```diff
@@ -104,14 +104,16 @@ class Search extends React.Component {
...
             <div className={classnames}>
-              <DebounceInput id="search-box"
-                              value={this.state.keywordSearched}
-                              debounceTimeout={300}
-                              type="search"
-                              onChange={(event) => this.handleInputChange(event)}
-                              placeholder="Search keywords, people, tags..."
-                              className="form-control"
-              />
+              <form action="for-search-key-to-show-on-mobile-keyboard" role="search">
```
mmmavis: Wrapping `<input type="search" ...>` with `<form action="not-blank">` does the trick

gvn:
Here's how this looks on my Nexus 5X in Chrome:

(When you tap the magnifying glass button it actually loads `/for-search-key-to-show-on-mobile-keyboard`. You will likely need to cancel that submit event when it fires.)

screenshot_20170829-153945
screenshot_20170829-153952

mmmavis:
@gvn thanks for catching this bug. Not sure why it worked on my iPhone in the first place... I probably had too many tabs open and was looking at the wrong tab when I did the test. I will update this PR.

mmmavis:
Okay I did some experiment and learned that if we call `event.preventDefault()` to cancel the form submit event the Search/Go key on keyboard then does nothing when you tap it (keyboard doesn't dismiss and page doesn't reload/change). If we really wanna have the Search/Go key we _can_ fix the form `action` so it points to a meaningful URI... but without `event.preventDefault()` tapping the Search/Go will still result a page reload. I don't think this is a good tradeoff... especially on mobile we would probably wanna avoid performing page reloading. 

I'm gonna close this PR and comment on the original ticket.

gvn:
Ok, yeah if you want to get rid of the keyboard you might have to unfocus the input after you prevent default.

Issue #672, opened 2017-08-23
Nav links should be in Open Sans


mmmavis:
/cc @kristinashu

kristinashu:
Looks good!

PR #678, opened 2017-08-30
Made nav links use Open Sans

mmmavis:
Related to #672

### DIFF
--- a/components/navbar/navbar.scss
+++ b/components/navbar/navbar.scss
@@ -13,7 +13,6 @@
     align-self: center;
 
     a {
-      font-family: "Open Sans";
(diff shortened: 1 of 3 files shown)


components/navbar/navbar.scss
```diff
@@ -13,6 +13,7 @@
     align-self: center;
 
     a {
+      font-family: "Open Sans";
```
mmmavis: is this enough or do we need to specify fallback font

gvn: You might want to add `"Helvetica", "Arial", sans-serif` as fallbacks.

mmmavis:
🎉

Issue #626, opened 2017-08-01
Create client side "edit profile" view


alanmoo:

xmatthewx:
comp: https://redpen.io/bk7c2a3cbaafdcd47a

profile view lives at: /user/[id]

 - [ ] display profile on success
 - [ ] display errors on failure
 - [ ] validate URLs
 - [ ] TBD country validation
 - [ ] TBD limit on bio char count
 - [ ] responsive
 - [ ] pulls name from google account
 - [ ] user can edit via link on their profile view
 - [ ] nobody else can edit
 - [ ] checklists can be existing mvp style 
 - [ ] all fields are optional

xmatthewx:
We likely wouldn't share with the public without Country validation if we believe we're going to do that. Hide behind feature flag.

mmmavis:
Work has been merged to the feature branch. Up for @xmatthewx do decide if we are good to close this ticket.

xmatthewx:
When I tap "Hi Matthew" I'm taken to /profile/me. View reports "Profile not Found" since I haven't made one yet.   

We probably need to do a migration pre-launch to establish default profiles for existing users. True? If sol can we file a ticket @gideonthomas? 

Should /profile/me redirect to the form when nothing is found? This would guard against errors. If so, we can file this as a follow up. 

cc @alanmoo

xmatthewx:
p.s. I'm looking at https://network-pulse-staging-pr-691.herokuapp.com/. Is this the latest branch?

xmatthewx:
When I view https://network-pulse-staging-pr-691.herokuapp.com/myprofile/

I see the form but no fields. 

<img width="614" alt="image" src="https://user-images.githubusercontent.com/53411/31180790-8490007a-a8ed-11e7-8503-380d432b6a7e.png">

xmatthewx:
Hold the phone. The API might be down. @alanmoo is investigating.

alanmoo:
https://network-pulse-staging-pr-691.herokuapp.com/myprofile works again

xmatthewx:
This is looking good. 

One thing I noticed ... but we can probably file as a follow up .... my profiled defaulted to name from google (that's good), but when i edit my profile, name field is blank. this is not terrible, but odd.

mmmavis:
@xmatthewx Let's file a new ticket for that. Since edit form is working... can we close this ticket now?

xmatthewx:
[...]

PR #679, opened 2017-08-30
Edit profile page UI only

mmmavis:
❗️ **This is a feature branch against another feature branch `edit-profile-page`.**

Related to #626 .
Mockup: https://redpen.io/bk7c2a3cbaafdcd47a

Instead of filing a huuuuuuuuuge PR I'm trying to tackle #626 in stages. This PR **only** covers the following

- added the route `/myprofile`
- added UI elements to the `/myprofile` page (no functionality, just making sure fields are there in the form)
- some CSS refactor and changes

Functionality bits like event handlers related to the form submission and form validations are not covered in this PR.

Note:
To log in locally... log in via http://test.example.com:8000/admin/login/?next=/admin/ instead. The "sign in" link on Pulse front end doesn't work locally due to some Google Client setup (I can't recall the actual reason. Will need to get insights from Pomax on this)

### DIFF
--- a/components/form-fields/help-types.jsx
+++ b/components/form-fields/help-types.jsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import DynamicCheckboxGroup from './dynamic-checkbox-group.jsx';
+import Service from '../../js/service';
+
+class HelpTypes extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { options: [] };
+  }
+  componentWillMount() {
+    Service.helpTypes.get().then(options => {
+      this.setState({ options: options.map(option => option.name) });
+    });
+  }
+  render() {
+    return <DynamicCheckboxGroup options={this.state.options} onChange={this.props.onChange} colNum={2} />;
+  }
+}
+
+export default HelpTypes;
--- a/components/form-fields/help-types.jsx
+++ b/components/form-fields/help-types.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import DynamicCheckboxGroup from './dynamic-checkbox-group.jsx';
 import Service from '../../js/service';
 
-class HelpTypes extends React.Component {
+class HelpTypesField extends React.Component {
   constructor(props) {
     super(props);
     this.state = { options: [] };
@@ -17,4 +17,4 @@ class HelpTypes extends React.Component {
   }
 }
 
-export default HelpTypes;
+export default HelpTypesField;
--- a/pages/add/add.scss
+++ b/pages/add/add.scss
@@ -1,187 +1,143 @@
 .add-page {
-  h1,
-  h2 {
-    color: $bikeshed-magenta;
-    margin-bottom: 0.5rem;
+  .posted-by {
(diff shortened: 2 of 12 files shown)


components/form-fields/help-types.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
mmmavis: `pages/add.jsx` and `pages/profile-edit.jsx` both have this field in the form. I extracted the code so this lives on its own file.

pages/add/add.scss
```diff
@@ -1,187 +1,143 @@
 .add-page {
```
mmmavis: A lot of the following rules has been 

- moved to `scss/forms.scss`
- or replaced by assigning boostrap classes

mmmavis: I think as long as the /add page still looks legit we are good.

pages/add/form/detail-info-fields.js
```diff
@@ -1,26 +1,11 @@
-import React, { Component } from 'react';
```
mmmavis: `pages/add.jsx` and `pages/profile-edit.jsx` both have this field in the form. I extracted the code so this lives on its own file.

components/form-fields/help-types.jsx
alanmoo: Nice

components/form-fields/help-types.jsx
```diff
@@ -0,0 +1,20 @@
...
+    Service.helpTypes.get().then(options => {
+      this.setState({ options: options.map(option => option.name) });
+    });
+  }
+  render() {
+    return <DynamicCheckboxGroup options={this.state.options} onChange={this.props.onChange} colNum={2} />;
+  }
+}
+
+export default HelpTypes;
```
alanmoo: Wanna call this `HelpTypesField` here since that's what you import it as elsewhere? Same for `IssuesField`?

pages/add/add.scss
```diff
@@ -1,187 +1,143 @@
...
-    margin-left: 0;
-  }
+  fieldset.creators {
+    .row {
+      vertical-align: middle;
+      margin: 0 0 0.5rem;
 
-  input {
-    @extend .form-check-input;
+      @media screen and (min-width: $bp-md) {
```
alanmoo: This feels a bit...weird. Is there a way to do this with bootstrap classes? Modifying the base grid classes has a tendency to go south, in my experience. Looks like you'd have to at least start with the `.multiple` element having `.row` as well, so you're not creating a new `.row` element with each creator.

mmmavis: Hey @alanmoo I agree with you but I think having a followup PR for this would be better. This chunk of CSS code is for the `/add` page and it was not newly introduced in this PR (the changes you saw are from indentation changes but the rules are the same).

alanmoo: Sounds good

mmmavis:
🎉 followup ticket created! https://github.com/mozilla/network-pulse/issues/686

PR #751, opened 2017-10-12
blur the input on Enter

### LINKED ISSUE #492: Seach keyboard on mobile should dismiss after hitting search/enter on keyboard

Pomax:
This uses an updated DebounceInput component, which as of v3 offers a direct link into the textfield that it uses so that we can blur it without needing to hack around it.

Fixes #492

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "postcss-cli-simple": "^1.0.3",
     "prop-types": "^15.5.6",
     "react": "^15.4.1",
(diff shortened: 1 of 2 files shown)


Pomax:
It looks like there's a fair number of errors on master that probably need a cleanup, too.

Issue #493, opened 2017-04-17
Form: show character count on the "title" field


kristinashu:
Title has a character count limit but we don't give users any feedback on this. (screen cap from production site)

<img width="758" alt="screen shot 2017-04-17 at 10 49 20 am" src="https://cloud.githubusercontent.com/assets/1068783/25098113/3cd0f576-235c-11e7-801c-45a6683a4f2c.png">

We should show the character count as they type (like on Twitter or in a google form). It is mocked up in redpen https://redpen.io/ov0d7f8ba9c437ee56

<img width="719" alt="screen shot 2017-04-17 at 10 50 18 am" src="https://cloud.githubusercontent.com/assets/1068783/25098106/38ef57a4-235c-11e7-8d49-3497357a0c89.png">

kristinashu:
Character count should show for both the Title field and Description field.

kristinashu:
Once we have added this to the Title field we should remove "Keep it simple. Name it, don't describe it." There is too much text on the form and the character count will keep people in check.

kristinashu:
@mmmavis do you think this would be a lot of work?

@xmatthewx can we put this into next heartbeat?

arush0311:
Can i take this?

mmmavis:
@arush0311 for sure! 😃 ✨ 

@kristinashu @arush0311 work needed to make this happen depends on how we want the counter to work. e.g., start showing the counter as soon as user starts typing in the field? or show counter after user's hit the "submit" button. Either way it will probably involve PRs on the form library we are using: [react-formbuilder](https://www.npmjs.com/package/react-form-builder) and since @Pomax wrote 99.99% of the code there I'll defer the question to him regarding how much work it would need to enable word/character counter.

kristinashu:
Yes, the counter should show as soon as user starts typing in the field (like Twitter).

kristinashu:
Realize twitter only shows the one number. If this is possible it would be great to do the same:

Before typing (see max number of character):
<img width="570" alt="screen shot 2017-08-01 at 1 42 13 pm" src="https://user-images.githubusercontent.com/1068783/28846261-9a0fb3e2-76bf-11e7-808f-60e3d308097a.png">

Starting typing (the number decreases as you type):
<img width="568" alt="screen shot 2017-08-01 at 1 42 30 pm" src="https://user-images.githubusercontent.com/1068783/28846269-a0193830-76bf-11e7-9b46-e94a84f72c4b.png">

[...]

PR #754, opened 2017-10-12
Add character limit to title

Pomax:
This PR adds in the character limit for "title" in a twitter style "show characters remaining" fashion:

typing

This closes https://github.com/mozilla/network-pulse/issues/493

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "react": "^15.4.1",
     "react-debounce-input": "^2.4.2",
     "react-dom": "^15.4.1",
-    "react-formbuilder": "^0.9.0",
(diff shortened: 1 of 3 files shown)


Pomax:
@kristinashu is this styling good enough or does the text need to include some kind of "Characters remaining: " prefix?

kristinashu:
Awesome, thanks @Pomax! We don't need a prefix, it's implied as soon as someone types.

It could just be your example but I want to making sure the character count for the title will still be 80 (as it is on production) not 140.

Pomax:
good catch, fixed to 80

PR #761, opened 2017-10-13
Applied character count to a few other fields

mmmavis:
Related to #493

### DIFF
--- a/pages/add/form/basic-info-fields.js
+++ b/pages/add/form/basic-info-fields.js
@@ -1,5 +1,10 @@
 import validator from './validator';
 
+let getRemainingCount = (charCount, charLimit) => {
+  // show a twitter-style "characters remainig" count
+  return charLimit - charCount;
+};
+
 module.exports = {
   title: { // required field
     type: `text`,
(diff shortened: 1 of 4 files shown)


pages/add/form/basic-info-fields.js
```diff
@@ -8,14 +13,10 @@ module.exports = {
     labelClassname: `required`,
     fieldClassname: `form-control`,
     validator: [
-      validator.emptyValueValidator(),
-      validator.maxLengthValidator(80)
```
Pomax: the charLimit does not do auto-validation right now, although I suppose it would make sense for it to do so. We cannot remove the validator functions here (yet).

Pomax: I've filed https://github.com/Pomax/react-formbuilder/issues/54 to address that but for now let's keep these validator functions so that we can be reasonably sure that all the validation we need actually kicks in.

mmmavis:
Depends on what comes out in https://github.com/mozilla/network-pulse/issues/769 ... I'll either have to update this PR or close it.

mmmavis:
I don't wanna have a stale PR so I'm closing this PR since 
1. followup tickets https://github.com/mozilla/network-pulse/issues/769 and https://github.com/mozilla/network-pulse/issues/770 are still in discussion
2. I won't be able to update this PR or work on those followup tickets until earliest the week of November 6.

Issue #752, opened 2017-10-12
Add background to image container for wide photos


xmatthewx:
Wide image from users don't fill the full height on the entry card. We get an awkward white space between thumbnail.

I propose we add a light gray background on the image's container, or something similar.
`.img-container { background: #eee; }` 

### Current

![image](https://user-images.githubusercontent.com/53411/31511885-1d61a2f8-af57-11e7-84b6-776758c9bc88.png)

This is not critical, but will add polish. 

@kristinashu - what do you think?

### Proposed

![image](https://user-images.githubusercontent.com/53411/31511906-33f7ce16-af57-11e7-9d96-c9c4c65bc7c2.png)

### Current

![image](https://user-images.githubusercontent.com/53411/31512076-b568b38e-af57-11e7-9a07-9c8432dec1de.png)

### Proposed A
Add gray background.

![image](https://user-images.githubusercontent.com/53411/31511936-5458b012-af57-11e7-82a6-05b4518af075.png)

### Proposed B
Collapse to height of the image. 

![image](https://user-images.githubusercontent.com/53411/31512061-a9afdce8-af57-11e7-9962-d00838929991.png)

kristinashu:
Yeah, if auto cropping is out of the questions then this is a good band-aid. 

Related to #721

PR #763, opened 2017-10-14
fix: Add background to image container for wide photos

aayushsanghavi:
Fixes: #752


Pomax:
I suspect we have a variable for that so we don't have to hardcode #EEE but I suspect @mmmavis knows this better than I do.

PR #768, opened 2017-10-16
ES6'd components

kamos1:
I've updated several components to use the ES6 class syntax. #477 @mmmavis


components/project-loader/project-loader.jsx
```diff
@@ -1,51 +1,53 @@
...
 
 const PROJECT_BATCH_SIZE = env.PROJECT_BATCH_SIZE;
 
-export default React.createClass({
-  getInitialState() {
-    return {
+class ProjectLoader extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
Pomax: keeping this the result of a function makes it possible to reset a component at any time, so it's worth keeping:
```
constructor(props) {
  super(props);
  this.state = this.getInitialState();
}
getInitialState() {
  return { 
    ...
  };
}
```
Keeping things separated this way allows for component resetting by calling `this.setState(this.getInitialState())`, inspecting the "what state should be at initi", etc.

components/project-loader/project-loader.jsx
```diff
@@ -1,51 +1,53 @@
...
     }
-  },
+  }
+
   componentWillReceiveProps(nextProps) {
     // Reset state before fetching data for the new params.
     // We want to keep existingPromise on record for fetchData
     // to handle it accoringly.
 
-    this.setState(this.getInitialState(), () => {
```
Pomax: we don't need to revise code here if we keep getInitialState responsible for "what the initial state should look like". Always good to keep code from repeating itself, especially with hardcoded values.

pages/add/add.jsx
```diff
@@ -14,9 +14,10 @@ import getHelpFields from './form/get-help-fields';
 const PRE_SUBMIT_LABEL = `Submit`;
 const SUBMITTING_LABEL = `Submitting...`;
 
-export default React.createClass({
-  getInitialState() {
-    return {
+class Add extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
Pomax: same comment as above for getInitialState()

pages/bookmarks.jsx
```diff
@@ -11,14 +12,16 @@ import utility from '../js/utility';
 import user from '../js/app-user';
 import pageSettings from '../js/app-page-settings';
 
-export default React.createClass({
-  getInitialState() {
-    return {
+class Bookmarks extends React.Component{
+  constructor(props) {
+    super(props);
+    this.state = {
```
Pomax: same comment as above for getInitialState

pages/bookmarks.jsx
```diff
@@ -86,12 +95,13 @@ export default React.createClass({
...
     });
-  },
+  }
+
   getContentForLoggedInUser() {
     let importHint;
 
     if (this.state.lsBookmarkedIds.length > 0) {
-      importHint = <span>Did you add favs while logged out? You can <button className="btn btn-link inline-link" onClick={evt=>this.handleImportBookmarksClick(evt)}>import favs</button> from this browser.</span>;
+      importHint = <span>Did you add favs while logged out? You can <button className="btn btn-link inline-link" onClick={(event) => this.handleImportBookmarksClick(event)}>import favs</button> from this browser.</span>;
```
Pomax: this did not need a refactor, so I'd strongly advise against changing this line as it isn't actually part of the work that this PR "is for": it does not change the behaviour of the code to be ES6 conformant, as it already was =)

If we want to change `evt` to `event` then we'll need to change that in a *lot* of other places, so let's roll back this particular change.

kamos1:
@Pomax I've added the changes you suggested

components/project-loader/project-loader.jsx
```diff
@@ -32,20 +40,12 @@ export default React.createClass({
...
       this.fetchData(this.props.bookmarkedOnly, nextProps);
     });
-  },
+  }
+
   createQueryParams(params) {
     let combinedParams = Object.assign({}, params);
 
     if (combinedParams.ids) {
-      // The `ids` query param is only used on the bookmarks(favs) page
```
Pomax: can we put this fairly large comment block back? If it requires an update, we should do that in a separate issue

Pomax:
Thanks @kamos1 - if you can restore that large comment block then this is probably ready for @mmmavis to give it a look-over, too.

kamos1:
@Pomax yeah, sorry about that. it should be good to go now.

Pomax:
lgtm - @mmmavis?

components/project-loader/project-loader.jsx
```diff
@@ -144,4 +156,10 @@ export default React.createClass({
       </div>
     );
   }
-});
+}
+
+ProjectLoader.propTypes = {
+  featured: PropTypes.string.isRequired
```
mmmavis: `featured` isn't a required `prop` so let's remove `isRequired`.

We should also check types for other `props` as well.

```js
ProjectLoader.propTypes = {
  featured: PropTypes.string,
  showCounter: PropTypes.bool,
  moderationState: PropTypes.object,
  bookmarkedOnly: PropTypes.bool,
  ids: PropTypes.array
};
```

pages/bookmarks.jsx
```diff
@@ -147,4 +165,15 @@ export default React.createClass({
       </div>
     );
   }
-});
+}
+
+Bookmarks.propTypes = {
```
mmmavis: only `this.props.router.location` is used in this component so I think we can change the this part to just

```js
Bookmarks.propTypes = {
  router: PropTypes.shape({
    location: PropTypes.object
  })
};
```

kamos1: @mmmavis I've fixed the proptypes.

PR #773, opened 2017-10-17
Profiles show created entries

### LINKED ISSUE #742: Add creator entries to Profile view
Currently, our new profiles only include entries `Submitted` by the person. Let's add things they `Created` as well. 

Note: MVP can just lump all entries together. Later, we'll want to separate by header or tabs. So let's proceed with that in mind.

@gideonthomas and @mmmavis – let's discuss:
 - how do we handle this on the frontend? is backend work required?
 - how big a body of work is this? feasible to land before mozfest? 

cc @alanmoo for awareness since profiles will be limited at mozfest without it.

alanmoo:
Closes #742 

I believe this is ready to review, staging API could use some data massaging but you can check http://localhost:3000/profile/5 to see the creators link working

### DIFF
--- a/components/project-card/meta/creators.jsx
+++ b/components/project-card/meta/creators.jsx
@@ -7,8 +7,22 @@ const Creators = (props) => {
 
   let classnames = classNames(`creator d-block open-sans`, props.className);
   let labelText = props.showLabelText ? `Created by ` : ``;
+  let joinedCreators;
 
-  return <p className="my-2"><small className={classnames}>{labelText}{props.creators.join(`, `)}</small></p>;
+  if (typeof props.creators[0] === `string` ) {
+    joinedCreators = props.creators.join(`, `);
+  } else {
+    let creators = [];
+    for (let creator of props.creators){
+      creators.push(<a href={creator.profile_id}>{creator.name}</a>);
+    }
+    // Because react doesn't like to render comma separated JSX objects easily
+    joinedCreators = creators.slice(1).reduce(function(prev, current) {
+      return prev.concat([`, `, current]);
+    }, [creators[0]]);
+
+  }
+  return <p className="my-2"><small className={classnames}>{labelText}{joinedCreators}</small></p>;
 };
 
 Creators.propTypes = {
--- a/components/project-card/meta/creators.jsx
+++ b/components/project-card/meta/creators.jsx
@@ -16,13 +16,13 @@ const Creators = (props) => {
     for (let creator of props.creators){
       // So that creators which are without a profile aren't links
       let url = null;
-      if(typeof creator.profile_id === "number") {
+      if(typeof creator.profile_id === `number`) {
         url = `/profile/${creator.profile_id}`;
       }
       creators.push(<a key={creators.length} href={url}>{creator.name}</a>);
     }
     // Because react doesn't like to render comma separated JSX objects easily
-    joinedCreators = creators.slice(1).reduce(function(prev, current) {
+    joinedCreators = creators.slice(1).reduce((prev, current)=>{
       return prev.concat([`, `, current]);
     }, [creators[0]]);
 
--- a/components/project-card/meta/creators.jsx
+++ b/components/project-card/meta/creators.jsx
@@ -7,31 +7,21 @@ const Creators = (props) => {
 
   let classnames = classNames(`creator d-block open-sans`, props.className);
   let labelText = props.showLabelText ? `Created by ` : ``;
-  let joinedCreators;
 
-  if (typeof props.creators[0] === `string` ) {
-    joinedCreators = props.creators.join(`, `);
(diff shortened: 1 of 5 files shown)


components/project-card/meta/creators.jsx
```diff
@@ -7,18 +7,39 @@ const Creators = (props) => {
 
   let classnames = classNames(`creator d-block open-sans`, props.className);
   let labelText = props.showLabelText ? `Created by ` : ``;
+  let joinedCreators;
 
-  return <p className="my-2"><small className={classnames}>{labelText}{props.creators.join(`, `)}</small></p>;
+  if (typeof props.creators[0] === `string` ) {
```
Pomax: What happens on a mixed list, where you get creators `['Pomax', 1, 25, 332]`?

components/project-card/meta/creators.jsx
```diff
@@ -7,18 +7,39 @@ const Creators = (props) => {
...
+      }
+      creators.push(<a key={creators.length} href={url}>{creator.name}</a>);
+    }
+    // Because react doesn't like to render comma separated JSX objects easily
+    joinedCreators = creators.slice(1).reduce((prev, current)=>{
+      return prev.concat([`, `, current]);
+    }, [creators[0]]);
+
+  }
+  return <p className="my-2"><small className={classnames}>{labelText}{joinedCreators}</small></p>;
```
Pomax: This might need a rewrite to simply do a map off of `props.creators` so that it just does "the right thing" based on the type, along the lines of
```
return props.creators.maps(data => {
  if (typeof data === `string`) {
    return <span className="creator" key={data}>{data}</span>;
  }
  let url = null;
  if(typeof data .profile_id === `number` && props.makeLink) {
    url = `/profile/${data.profile_id}`;
  }
  return <a className="creator" key={data.profile_id} href={url}>{data.name}</a>);
});
```
And then have CSS automatically inject that comma using a `creator + creator:before { content: ", "; }` rule.

components/project-card/meta/creators.jsx
alanmoo: You wouldn't, you'd get either creators from `creators` (array of strings) or `related_creators`(array of objects, which we should be using everywhere from now on). I put this code in before I built in the `showLink` prop, so with that I think I can safely remove this, and simply define the propType of `creators` as an array of objects.

components/project-card/meta/creators.jsx
```diff
@@ -7,18 +7,39 @@ const Creators = (props) => {
 
   let classnames = classNames(`creator d-block open-sans`, props.className);
   let labelText = props.showLabelText ? `Created by ` : ``;
+  let joinedCreators;
 
-  return <p className="my-2"><small className={classnames}>{labelText}{props.creators.join(`, `)}</small></p>;
+  if (typeof props.creators[0] === `string` ) {
+    joinedCreators = props.creators.join(`, `);
+  } else {
+    let creators = [];
```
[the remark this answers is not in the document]
alanmoo: Oh, the key can be the creator name? I was stuck because the id could be null.

components/project-card/meta/creators.jsx
```diff
@@ -7,18 +7,39 @@ const Creators = (props) => {
...
+    for (let creator of props.creators){
+      // So that creators which are without a profile aren't links
+      let url = null;
+      if(typeof creator.profile_id === `number` && props.makeLink) {
+        url = `/profile/${creator.profile_id}`;
+      }
+      creators.push(<a key={creators.length} href={url}>{creator.name}</a>);
+    }
+    // Because react doesn't like to render comma separated JSX objects easily
```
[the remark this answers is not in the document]
alanmoo: Because it was rendering as `[Object object], [Object object]` or something like that. This could be my JSX is rusty and I can't remember the rules around rendering things- maybe I shouldn't be using an array of JSX nodes?

pages/profile.jsx
```diff
@@ -15,19 +15,27 @@ class Profile extends React.Component {
...
-      />
-    </div>;
+    return <div className="row">
+      <div className="col">
+        <h2 className="h4">
+          {label}
+        </h2>
+      </div>
+      <div className="col-12">
+        {/* TODO:FIXME: for now let's just render all entries at once. (no 'view more' button) I will file another PR to refine this. */}
```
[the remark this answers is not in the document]
alanmoo: This isn't my code. @mmmavis is a ticket open for this?

components/project-card/meta/creators.jsx
alanmoo: CSS didn't feel like a good idea for separation of concerns reasons. I ended up using `reduce` to squash them together and that works well enough.

alanmoo:
Pomax R+'d in slack, so wielding my admin power to merge

Issue #741, opened 2017-10-05
[name] displayed for loggedin user should be that person' custom name (if it exists)


mmmavis:
Green circled text should all be `mmm` (my custom name)

![screen shot 2017-10-05 at 3 35 08 pm copy](https://user-images.githubusercontent.com/2896608/31256187-c9de35f4-a9e5-11e7-836e-e4982df8ce1a.jpg)

and

![screen shot 2017-10-05 at 3 35 21 pm](https://user-images.githubusercontent.com/2896608/31256169-aa3c5e24-a9e5-11e7-990b-16cc4250065f.jpg)

Related backend ticket: https://github.com/mozilla/network-pulse-api/issues/229 so we can easily get a user's `name` by calling `/userstatus` endpoint

xmatthewx:
verified on staging! hurrah. closing.

Issue #727, opened 2017-10-04
Profile edit form should inlude default name


xmatthewx:
As an existing user, on staging, my profile defaulted to my name from google. ✔️ That's good. But,  when i went to edit my profile, the name field was blank. This required me to retype my name. I didn't notice, and my submit failed. Not a huge deal, but it adds odd friction for our users.

mmmavis:
This is somewhat blocked by https://github.com/mozilla/network-pulse-api/issues/229 (so that we can rely on the API response to determine what "name" to show on front end). I could get a interim solution in first but not sure if I should as it introduces some complexity to the codebase.

xmatthewx:
verified on staging!! closing.

PR #780, opened 2017-10-20
Show user's custom name where applicable

mmmavis:
- (first commit) Related to #741 - display user's custom name if it exists. otherwise fall back to user's username
- (second commit) Related to #727 - profile edit form should include default name


mmmavis:
yay @gideonthomas thanks for reviewing!

PR #781, opened 2017-10-20
More es6ing

kamos1:
updated the profile-edit component to an es6 class #477


kamos1:
@Pomax @mmmavis updated another component #477

PR #811, opened 2017-10-23
update profile edit component

kamos1:
This PR updates the profile-edit component to es6 for #477 @mmmavis


alanmoo:
Thanks for the contribution! Looks like you've got a few CI errors to fix, and I wanted to let you know we're going to be waiting until after Mozfest this weekend to land any more non-bugfix PRs.

In the meantime, I know we had some ES6 issues related to `this` binding recently, so before anyone merges this in, let's double check that we don't have that problem again.

PR #815, opened 2017-10-23
handle incomplete input for autocomplete fields

### LINKED ISSUE #715: "tags" field does not commit a tag to the list if you just paste it in
if you paste, rather than type, a tag, it will not show up in the list of tags until you hit `,` or tab or enter. As such, the following leads to an entry without any tags at all:

copy `a,b,c,d` to the clipboard, create a new entry:

title: whatever
content_url: http://whatever.com
tags: paste from clipboard

then hit "submit"

This will send an entry without any tag data, as "nothing was typed" in the tags field (it does not seem paste aware yet)

### LINKED ISSUE #816: tags/creators are lost if the user does not explicitly auto-complete them.
It's possible for user input for tags and creators to get lost, even though from a user's perspective that should clearly get recorded, in the case of the user clicking away from an autocomplete field, rather than tabbing to the next one or accepting an autosuggestion from the input element.

This makes it possible for entries with single tags to get submitted without any tags at all if the user clicks away to the next field, or if they write an exact creator name but do not first manually accept a suggestion.

Pomax:
This unifies the tag/creators using a shared class that takes care of handling "incomplete data" (e.g. not autocomplete) when a users clicks away (not tabs away) from an autocomplete field. This does different things based on the subclass:

- Tags: autocomplete the text the user had written (turning it into a real tag)
- Creators: only accept the input as a creator if there is an exact match with a record found in the suggestion list. Otherwise, discard the input entirely.

Fixes #816, unblocks #715

To test, run `npm start` and on the http://localhost:3000/add page;

For creators, input a creator string that will match a suggestion, then click somewhere outside the input field. It should get auto-converted to the matching suggestion. Try a second time with nonsense data that will yield no suggestions, and click outside the input field. The input should get thrown away.

For tags, input any string and then click outside the input field. The text should immediately get turned into a real tag.


pages/add/form/creators.js
```diff
@@ -1,21 +1,26 @@
...
-  }
-  getInitialState() {
+export default class Creators extends AutoCompleteInput {
+  toSchema(field) {
+    // Model data thusly:
+    //
+    // "related_creators": [{
+    //   "creator_id": number | <id of the creator autocompleted. Will be null if the creator was not autocompleted and was manually typed in>,
+    //   "name": string | <name of the creator typed in, not autocompleted. Will be null if the creator was autocompleted>
+    // }, ...]
```
alanmoo: Where does `profile_id` fit in here?

pages/add/form/creators.js
```diff
@@ -30,80 +35,17 @@ export default class Creators extends Component {
...
-  fixCreatorLettercase(creator) {
-    for (let t of this.state.suggestions) {
-      // We don't want to have duplicated creators in the database.
-      // If the user-defined creator is a known creator,
-      // fix the lettercase so it mataches the known creator we have in the database.
-      if (t.name.toLowerCase() === creator.name.toLowerCase()) {
-        creator = t;
-        break;
+  handlePendingInput(input) {
+    // If the creator the user typed matches a known creator suggestion (ignoring
```
alanmoo: This isn't the behavior I get. Even when it's an exact match for the creator name, it gets discarded when I click away. (I should also point out that this differs from the delimiter case- even if there's exactly one match, we create a new creator unless the user specifically selects the match from the dropdown- Also worth creating a new ticket?)

pages/add/form/creators.js
Pomax: I don't know, this refactor was only a refactor - probably good to ask @gideonthomas and @gvn whether that id is still necessary

pages/add/form/auto-complete-input.js
```diff
@@ -0,0 +1,116 @@
+import React, { Component } from 'react';
+import ReactTags from 'react-tag-autocomplete';
+import Service from '../../../js/service';
+import { DELIMITERS } from './tag-delimiters';
+
+function findReactComponent(dom) {
+  for (var key in dom) {
+    if (key.startsWith("__reactInternalInstance$")) {
```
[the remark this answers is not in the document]
Pomax: good flag - this is scaffolding that should have been removed.

pages/add/form/creators.js
alanmoo: It is, see line 29

pages/add/form/creators.js
Pomax: Turns out this happens because the profile marker is a bit of unicode string prepended to the user name.

PR #824, opened 2017-10-24
Add a codeowners file.

alanmoo:
We can adjust this as necessary, but we need something for now

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @alanmoo @xmatthewx


Pomax:
I am incredibly curious how we plan to keep this file up to date, which makes me highly apprehensive of landing a file like this.

alanmoo:
But the alternative is that contributor PRs fall into a black hole unless someone notices them. We could make everyone get a notification (via github groups), but that seems like a bunch of unnecessary noise.

mmmavis:
Maybe we can triage contributor PRs during our daily stands?

alanmoo:
For projects we're actively working on, that's great. My concern is about when we're not doing that, but still have it in maintenance mode. By adding a `codeowners` file there's at least a chance that someone will see it. We could even just have it request a PR from the whole engineering team (using github groups)

xmatthewx:
We can add more names if that's the issue. Any pathway for contacting us makes this more accessible than no pathway.

Issue #807, opened 2017-10-23
Show Favs help text for logged-in users


xmatthewx:
This view only shows the empty view tip to save your favs if a user is not logged in. Logged in users just get an empty view if they have no favs. 

![image](https://user-images.githubusercontent.com/53411/31903459-270d66a0-b7f6-11e7-87ee-e0defe25e564.png)

alanmoo:
Just to be clear- the request here is that "the content of the screenshot should show up if a user is logged in but doesn't have any favs", is that correct?

xmatthewx:
Yes, everyone should see this help text if they have no favs, regardless of auth state.

xmatthewx:
perfecto

PR #835, opened 2017-11-08
Show bookmark prompt for loggedin user

mmmavis:
Related to #807


pages/bookmarks.jsx
```diff
@@ -73,6 +74,12 @@ class Bookmarks extends React.Component{
     });
   }
 
+  showBookmarkPrompt(show = false) {
```
gvn: This is kind of confusing to me, especially with logic inside of the method that might actually hide the bookmark?

I usually make methods like this not have any logic and just do what they are called (eg: `showBookmarkPrompt` and `hideBookmarkPrompt`) and then put logic elsewhere.

Issue #731, opened 2017-10-05
Implement nav ui alignment 


kristinashu:
From alignment work in mozilla/network#582 and to align better with new brand, mozilla.org, Network site, and IRH sites.

https://redpen.io/fv1d606e85be529df4

![pulse home 2x](https://user-images.githubusercontent.com/1068783/31301808-b3d6913a-aab1-11e7-95f9-dccaf522372b.png)

### Updates include:
- keep logo the same
- update menu items Open Sans Bold 16px
- search icon the same
- left align nav items and seach
- gradient from #FF506E to #A56EFF | 4px high
- Signin / Signup or Username right aligned #A668FF
- hover state for nav items and Username is black underline 2px high
- Add entry CTA button has hover state and styling similar to https://network.mozilla.org/ but 40x40px, #A668FF with black stock and shadow, icon [icon-add.svg.zip](https://github.com/mozilla/network-pulse/files/1360214/icon-add.svg.zip)
- total nav bar height is 88px
- mobile icons [Icons.zip](https://github.com/mozilla/network-pulse/files/1364105/Icons.zip)

### Mobile:
Similar to https://network.mozilla.org/ 

#### Medium width
<img width="582" alt="screen shot 2017-10-06 at 4 27 19 pm" src="https://user-images.githubusercontent.com/1068783/31302009-467982e4-aab3-11e7-8172-278fddea17b5.png">

#### Narrow width
<img width="770" alt="screen shot 2017-10-06 at 4 41 10 pm" src="https://user-images.githubusercontent.com/1068783/31302266-3ccbcd22-aab5-11e7-8c9e-af5125512fc0.png">

cc @mmmavis

xmatthewx:
@mmmavis – I expect that this work will take a bit of time and focus. In terms of priorities, let's stick to Profile related things first. I discussed with Jesse, and we want to make sure we ship that stuff well first, and then add this UI polish. I don't want you to feel overwhelmed given your other MozFest duties. 

Make sense? Let me know if you have any questions or want to discuss. 

When ready, we should also connect with Gavin to align how we handle the hamburger yum yum. Again, I think you don't need to worry about this unless you're blocked on profile related things. cc @alanmoo

mmmavis:
@kristinashu are we removing "Featured" from nav bar?

kristinashu:
No, you can keep Featured.

kristinashu:
[...]

PR #845, opened 2017-11-16
Nav bar tweaks

mmmavis:
Related to #731

### Addressed the following items on https://github.com/mozilla/network-pulse/issues/731#issuecomment-344330899

> - reduce height of the entire header (about from 220px to 180px)
> - make '+' icon hover state smooth like on Foundation site, rather than sudden jump
> - selected state for menu item should be underlined
> - link the little 'm' to the Feature page (like the full logo)

### Didn't get addressed

> - here's the logo in black pulse-logo.svg.zip

Not sure what this file is for. Might be for logo hover state? I checked a few Mozilla sites and none of them inverses logo colour on hover state. --- let's discuss this further in a followup ticket.

> Bug: (maybe open a new ticket for this?)
> If open the menu at a narrow screen width and then make it wide, the menu looks like this (seems ok on Foundation site):

This is probably not worth fixing as it is a rare case and there's probably no "clean" way of fixing it (without JS or duplicated DOM elements).

### DIFF
--- a/components/navbar/navbar.jsx
+++ b/components/navbar/navbar.jsx
@@ -131,7 +131,7 @@ class NavBar extends React.Component {
               </IndexLink>
               { this.renderNavList() }
             </div>
-            <div className="pinned col-12 col-lg-3">
+            <div className="pinned col-6 col-lg-3">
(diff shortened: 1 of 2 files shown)


mmmavis:
🎉

Issue #806, opened 2017-10-23
Staging QA: "Add Page" form error messages are hard to notice.


flukeout:
Here's a cap of a field with an 'error state'. I think it's too subtle given that we already use red for headings. Also on "mobile/narrow" mode, we don't scroll the page up when you have errors, so the user is left having to scroll up to notice these.

Some ideas..
* Add red border to field
* Red background, white text on the indicator
* Move indicator closer to field
* Add little arrow pointing up

![image](https://user-images.githubusercontent.com/25212/31903374-c7d07ec0-b7dc-11e7-83ae-2a3bb354aadc.png)

kristinashu:
Yeah let's do what was in the comps! 

- Red border around the field
- Indicator close to field
- X next to text

<img width="685" alt="screen shot 2017-11-08 at 3 19 10 pm" src="https://user-images.githubusercontent.com/1068783/32580014-46d356b2-c498-11e7-9dab-e71cc7578f48.png">

mmmavis:
Works on staging!

PR #853, opened 2017-11-18
Form error styling tweaks

mmmavis:
Related to #806 

---

@kristinashu 

screen shot 2017-11-17 at 4 44 33 pm

We can't smartly and easily position error message and character counter on the same line because each of them belongs to a different parent element (green box and blue box). Do you think removing "x This field cannot be left blank" and just showing a red "0/80" is enough?

### DIFF
--- a/pages/add/add.scss
+++ b/pages/add/add.scss
@@ -5,7 +5,7 @@
 
   fieldset.title {
     .error {
-      margin-right: 50px;
+      margin-right: 3rem;
     }
   }
 
--- a/pages/add/add.scss
+++ b/pages/add/add.scss
@@ -3,6 +3,12 @@
     margin-top: $form-field-vertical-margin / 2;
   }
 
+  fieldset.title {
+    .error {
+      margin-right: 50px;
+    }
+  }
+
   fieldset.published_by_creator {
     input {
       margin-right: $form-check-input-margin-x;
--- a/scss/main.scss
+++ b/scss/main.scss
@@ -110,13 +110,6 @@ hr.hr-gradient {
 
 form {
   font-family: inherit;
-
-  .inline.error {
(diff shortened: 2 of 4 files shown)


xmatthewx:
We should not remove error message, even if we can't improve the position. 

I wonder if you can put `position:absolute` on `.char-limit` to pull it out of the flow. I did a quick test and it doesn't ruin any margins or padding and it doesn't collide with the error message even if it is super duper long.

mmmavis:
@xmatthewx Alright. I try to avoid using absolute positioning as it usually requires more manual fixes across different screen sizes and devices. Let's try this approach. Maybe it won't be that bad for this case!

scss/main.scss
```diff
@@ -110,13 +110,6 @@ hr.hr-gradient {
 
 form {
   font-family: inherit;
-
-  .inline.error {
```
mmmavis: moved this to `forms.scss` since this is for overriding rules from `react-formbuilder`

pages/add/add.scss
```diff
@@ -3,6 +3,12 @@
     margin-top: $form-field-vertical-margin / 2;
   }
 
+  fieldset.title {
+    .error {
+      margin-right: 50px;
```
alanmoo: Let's use `em` here, since this margin is specifically to pad for text on the right side. `3em` looks good.

mmmavis: Do we have convention about when to use `px`, `rem`, and `em`? I've been using `rem` for mostly font sizes and `px` for spacing. I almost never used `em`...

alanmoo: I don't think we do, though we probably should (cc @gvn). You could go with rem here, based on the size of the text being used for the character count.

Issue #470, opened 2017-04-07
Upgrade react-router to v4


mmmavis:
https://reacttraining.com/react-router/

https://github.com/ReactTraining/react-router/releases/
https://github.com/ReactTraining/react-router/releases/tag/v4.0.0-0
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/migrating.md

http://rants.broonix.ca/upgrading-to-react-router-v4/

---

Serving side rendering doc
- https://reacttraining.com/react-router/web/guides/server-rendering

mmmavis:
Now might be a good time to work on this ticket. Moving it to the current milestone first so I can revisit next week.

mmmavis:
This works good on staging.

PR #872, opened 2017-11-29
Update router-router to v4.2.0

mmmavis:
Related to #470

@gvn `react-router` v4 is a lot different than v3 so that explains why many code changes are included in this PR. I think you can just use the review app to test things: https://network-pulse-staging-pr-872.herokuapp.com/

Main things to verify
- pages get served properly (project list gets loaded, no console error or server error)
- log in/out works
- form submission works 

I think as long as the above things work this PR is good to merge. (We can file followup ticket if small bugs are discovered)


mmmavis:
🎉

Issue #867, opened 2017-11-22
Update thanks message for unmoderated submissions


xmatthewx:
When a User submits a new entry, we display a thank you message. We have one version for non-moderators (thanks w/o published entry), one version for moderators (thanks w/ published entry).

For non-moderators, we recently updated their thank you message to point them to their profile and encourage them to edit it. 

For staff/moderators, we should do something similar. 

Proposed implementation for staff/moderators:
 - [ ] shorten header to "Thanks!" 
 - [ ] add paragraph text with link to profile 
 - [ ] add button with link to edit profile
 - [ ] no other changes to submission flow or display of entry

![image](https://user-images.githubusercontent.com/53411/33140754-e4c31326-cf7e-11e7-8d60-a0b8570bdf50.png)

Their new entry appears below this thank you message (no change from current implementation).

mmmavis:
@xmatthewx It seems like the `/api/pulse/profiles/me/` endpoint doesn't include projects that haven't been approved. So either we update the backend or we change the "submission was added to your profile" line to something else.

mmmavis:
Oops I guess I read the ticket description wrong - was the proposed mockup for staff/moderator?

xmatthewx:
Yes. I edited my initial comment to try and make it more clear. Slack me if this is still confusing. I can't login to staging right now to add more screenshots.

mmmavis:
This is the screen a moderator would get after posting an entry.

![screenshot-2017-11-29 yay xmas - mozilla network pulse](https://user-images.githubusercontent.com/2896608/33400824-1853df1c-d50c-11e7-93f8-27a70e7a91cc.jpg)

@xmatthewx is the plan to replace everything (Thanks for Submitting + entry details) with the messages and CTA you mentioned in the ticket description?

xmatthewx:
> is the plan to replace everything (Thanks for Submitting + entry details) with the messages and CTA you mentioned in the ticket description?

No. We should only replace "Thanks for Submitting" with the new copy and CTA. We should keep the entry details. 

Note: the /submitted page for non-moderators should also show button that says Edit Profile. Gideon and I implemented that at MozFest.

mmmavis:
> Note: the /submitted page for non-moderators should also show button that says Edit Profile. Gideon and I implemented that at MozFest.

Nice catch. Yea the button is still there. It's just when I took the screencap I cheated by editing some code so I could skip the entry submission part to get to that screen. 😂

[...]

PR #884, opened 2017-12-05
Purple notification bar + some code cleanup

mmmavis:
Related to #867 

## Example of the screen a moderator would see after submitting an entry (entry gets auto approved right away)

https://network-pulse-staging-pr-884.herokuapp.com/entry/373?justPostedByUser=true (note that the query `justPostedByUser=true` should be clear after the page gets loaded)

screenshot-2017-12-4 finger heart love - mozilla network pulse

## Example of the screen a non moderator would see after submitting an entry (entry is in "pending" state)

https://network-pulse-staging-pr-884.herokuapp.com/submitted

screenshot-2017-12-4 thank you - mozilla network pulse

### DIFF
--- a/components/notification-bar/notification-bar.scss
+++ b/components/notification-bar/notification-bar.scss
@@ -1,0 +1,7 @@
+.notification-bar {
+  @extend .open-sans;
+
+  font-weight: 600;
+  background: rgba($pulse-purple, 0.2);
+  padding: 1rem;
+}
--- a/components/notification-bar/notification-bar.scss
+++ b/components/notification-bar/notification-bar.scss
@@ -3,5 +3,4 @@
 
   font-weight: 600;
   background: rgba($pulse-purple, 0.2);
-  padding: 1rem;
 }
--- a/pages/add/submitted.jsx
+++ b/pages/add/submitted.jsx
@@ -1,65 +1,20 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
-import qs from 'qs';
 import { Helmet } from "react-helmet";
 import HintMessage from '../../components/hint-message/hint-message.jsx';
 
-const HEADER_TEXT = `Thanks!`;
-
-export default class Submitted extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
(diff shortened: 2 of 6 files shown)


pages/add/submitted.jsx
```diff
@@ -1,65 +1,20 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
-import qs from 'qs';
```
mmmavis: This `Submitted` component should really be a stateless static component.

pages/entry.jsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 import { Helmet } from "react-helmet";
+import { Link } from 'react-router-dom';
 import qs from 'qs';
 import LoadingNotice from '../components/loading-notice.jsx';
-import ProjectCardDetialed from '../components/project-card/project-card-detailed.jsx';
+import NotificationBar from '../components/notification-bar/notification-bar.jsx';
+import ProjectCardDetailed from '../components/project-card/project-card-detailed.jsx';
```
mmmavis: fixed typo in "detail"

components/notification-bar/notification-bar.scss
```diff
@@ -0,0 +1,7 @@
+.notification-bar {
+  @extend .open-sans;
+
+  font-weight: 600;
+  background: rgba($pulse-purple, 0.2);
+  padding: 1rem;
```
gvn: Is there a Bootstrap padding class that can be used instead?

mmmavis: you mean do `@extend .p-3` here or add `p-3` as class to the notification component?

gvn: Just adding a class to the component...

mmmavis: @gvn this component will be used across the site. I wanna make sure it has consistent/fixed padding so I'd like keeping the padding rule here.

gvn: If the class is in the React component definition it should be consistent everywhere it's used.

Issue #805, opened 2017-10-23
QA: Increase social media icon size on mobile


xmatthewx:
On mobile, social icons are pretty tiny. Should have minimum of 40px touch target.

mmmavis:
Do we want to 1) make those icon bigger (40px wide) or 2) add spacing in between icons?

kristinashu:
Icon size is ok, just made the hint area bigger please.

xmatthewx:
this  screenshot shows the relative size of icons on profile versus entry. the profile ones still feel small to me. we can live with this, nobody will be injured, but seems like a good opportunity to align styles.  

<img width="219" alt="image" src="https://user-images.githubusercontent.com/53411/32967259-ef2f63aa-cbaa-11e7-849f-978bcc1eb8a7.png">

kristinashu:
Oh I didn't realize you where talking about the links in profiles... I'll take a look.

mmmavis:
The fix to increase hint area has been merge to `master` and pushed to staging. If we want to change icons on profile so they align with icons on entry we should file a new ticket!

mmmavis:
@kristinashu I saw you moved this to the "Up Next" pipeline. Give me a ping when this is ready for implementation again!

kristinashu:
@mmmavis could you please make the profile icons (Twitter, LinkedIn, and GitHub) the same height as the post icons are (heart and Twitter share). And then make all their hit areas 40x40px.

mmmavis:
@kristinashu 

> Could you please make the profile icons (Twitter, LinkedIn, and GitHub) the same height as the post icons are (heart and Twitter share). And then make all their hit areas 40x40px.

Making the hit areas exactly 40x40px would be a bit complicated as each icon has different dimensions plus we have consider padding and margin. I'll try to balance things out and make the hit areas as close to 40x40px as I can.

mmmavis:
@kristinashu ah forgot to ask -- Are these changes for mobile only? (as described in Matthew's ticket description)

kristinashu:
Roughly around 40x40px is great. These updates can be for both mobile and desktop.

himanish-star:
@mmmavis , is this issue still available. Can I fix this ?

mmmavis:
@himanish-star thanks for offering help. I have not yet had time to tackle this. Would you be interested in taking on this ticket? 😃

himanish-star:
@mmmavis , Yeah sure

[...]

PR #907, opened 2018-01-12
Social media Icon sizes have been increased to 40px

himanish-star:
Related issue: #805 
Summary : @mmmavis , I have made the changes and as the icons were `font-awesome` icons so I only increased their `font-size` to `40px`.
Here is a screenshot of the changes made
screenshot from 2018-01-12 18-00-24
 
and for desktop view
screenshot from 2018-01-12 18-00-47

### DIFF
--- a/components/bio/bio.jsx
+++ b/components/bio/bio.jsx
@@ -42,7 +42,7 @@ class Bio extends React.Component {
       let link = this.props[type];
       if (!link) { return; }
 
-      let classname = classNames(`d-inline-block social-media px-2 px-sm-0`, {
+      let classname = classNames(`d-inline-block social-media`, {
         "mr-sm-3" : i !== list.length-1
       });
 
--- a/components/bio/bio.scss
+++ b/components/bio/bio.scss
@@ -13,6 +13,16 @@
     }
   }
 
+  .social-media {
+    width: 40px;
+    text-align: center;
+
+    @media screen and (min-width: $bp-md) {
+      width: auto;
+      text-align: left;
+    }
+  }
+
   .name {
     font-size: 1.75rem;
     font-weight: 500;


components/bio/bio.scss
```diff
@@ -13,6 +13,12 @@
     }
   }
 
+  @media screen and (max-width: 500px) {
```
mmmavis: For this codebase we want to use "mobile first" approach so instead of using `max-width`, we want `min-width` here instead.

Learn more about `mobile first` CSS approach: https://zellwk.com/blog/how-to-write-mobile-first-css/

mmmavis: Try to stick with breakpoint variables defined in `/scss/overrides/_variables.scss` instead of using an arbitrary number (e.g., 500px).

components/bio/bio.scss
```diff
@@ -13,6 +13,12 @@
...
+    .social-media span {
+      font-size: 40px;
```
mmmavis: From https://github.com/mozilla/network-pulse/issues/805#issuecomment-343208753 and https://github.com/mozilla/network-pulse/issues/805#issuecomment-349737177
> Icon size is ok, just made the hint area bigger please.

> @mmmavis could you please make the profile icons (Twitter, LinkedIn, and GitHub) the same height as the post icons are (heart and Twitter share). And then make all their hit areas 40x40px.

We want to make the hint areas bigger while keeping the icon size (font size in this case) the same. So increasing the font-size to 40px is not what we want there 😬 .

himanish-star: @mmmavis , sure I'll make the changes

himanish-star: Could you explain hit area ? Do you mean padding ? And do I have to change the hit area to 40px in mobile view or normal view.

himanish-star:
image
image
@mmmavis , sorry for the delay. Above are the preview images.

himanish-star:
@mmmavis , I knew it that I was going wrong as the desktop view due to my changes look horrible :disappointed: :. I knew setting min-width of 576px won't work but it didn't strike me that I should use $bp-md. I'll make the changes soon

components/bio/bio.jsx
```diff
@@ -42,7 +42,7 @@ class Bio extends React.Component {
       let link = this.props[type];
       if (!link) { return; }
 
-      let classname = classNames(`d-inline-block social-media px-2 px-sm-0`, {
+      let classname = classNames(`d-inline-block social-media`, {
         "mr-sm-3" : i !== list.length-1
```
himanish-star: Shall I also remove this

himanish-star:
Changes made @mmmavis

components/bio/bio.jsx
mmmavis: um I think margins are okay!

himanish-star:
>  Let me know if any part of my suggested solution doesn't make sense. :grin:
Thanks, everything makes sense now

Issue #910, opened 2018-01-19
Clean up Help URLs


xmatthewx:
The /help/ URLs are odd, and don't match how we handle /issue/ URLs. Let's iterate. 

 - Current: `https://www.mozillapulse.org/help/Plan%20%26%20organize` 
 - Expected: `https://www.mozillapulse.org/help/test-&-feedback`

Here's a staging entry with all help categories: https://network-pulse-staging.herokuapp.com/entry/378 

 - [ ] replace space with hyphen
 - [ ] lowercase
 - [ ] `&` unencoded
   - For issues, we use `and` instead of `&`. I'm fine with either, and think `&` is best, but I don't know if it's a bad practice to use that for some reason. 

All URLs:
/help/attend
/help/create-content
/help/code
/help/design
/help/fundraise
/help/join-community
/help/localize-&-translate
/help/mentor
/help/plan-&-organize
/help/promote
/help/take-action
/help/test-&-feedback
/help/write-documentation

cc @alanmoo to triage

shikhar-scs:
@xmatthewx I would like to take up this issue. May I ?

alanmoo:
Sure! Though `&` will need to be replaced by `and` because that character must be perecent encoded in URLs.

shikhar-scs:
@alanmoo can you help me on where to make these exact changes?

alanmoo:
These aren’t just string changes; I believe these help categories are stored in the database, so you’d probably need some logic around encoding URLs for the browser that act differently than XHR requests to the API. Though thinking about it again, that could result in some inconsistencies depending on the specific tags used (particularly ones with `&`). @gideonthomas do you have any thoughts?

shikhar-scs:
@gideonthomas

alanmoo:
He’ll see this Monday, Eastern Standard Time.

shikhar-scs:
oops ok :+1:

gideonthomas:
@alanmoo this is really a client-side routing issue! All we store in the database is the name of the help type. All the routing including transforming the help type name into a route is done client-side via react-router. 
Right now the front-end just `encodes/decodes` the name of the help type to figure out the route. You can probably change that to use some js slug library instead but would need to find out how to retrieve the non-slugified version once you hit the route so that you can retrieve data from the api.

shikhar-scs:
>  All the routing including transforming the help type name into a route is done client-side via react-router.

[...]

PR #919, opened 2018-02-07
Prettify helpType slug

mmmavis:
Related to #910

### DIFF
--- a/js/utility.js
+++ b/js/utility.js
@@ -36,6 +36,24 @@ const Utils = {
     }
 
     return ``;
+  },
+
+  getSlugFromHelpTag(helpTag) {
+    // this function turns help type into a prettier string so we can use it as slug
+    // e.g., "Plan & organize" becomes "plan-and-organize"
+
+    return helpTag.split(` `)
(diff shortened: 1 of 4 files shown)


js/utility.js
```diff
@@ -36,6 +36,24 @@ const Utils = {
...
+      .map(word => encodeURIComponent(word.replace(/&/g, `and`)).toLowerCase())
+      .join(`-`);
+  },
+
+  getHelpTagFromSlug(slug) {
+    // turns help type slug into a string Pulse API recognizes as helpType
+    // e.g., "plan-and-organize" becomes "plan & organize"
+
+    return slug.split(`-`)
+      .map(word => decodeURIComponent(word.replace(/and/g, `&`)))
```
Pomax: this will turn "grand" into "gr&", you probably want `word.replace(/\band\b/g, `&`)` here which only replaces `and` if it has word boundaries on either side (such as start/end of phrase, whitespace, etc).

mmmavis: yay regex master to rescue!

mmmavis:
🎉

PR #933, opened 2018-03-08
Added default profile pic

### LINKED ISSUE #928: Add a fallback image for users without an avatar
For users who do not have not uploaded an avatar, let's set the image below to display on their profile. 

See https://github.com/mozilla/network-pulse/issues/881 for prior discussion.

profile-placeholder

mmmavis:
Fixes #928


gvn:
Seems like that'd be better as a PNG or SVG, but it's fine...

Issue #955, opened 2018-05-07
Follow up on new Profile Favs


xmatthewx:
I see three things worth fixing. I don't think these need to block shipping the feature, and can be follow up so we can get the feature live. @mmmavis  

 - [x] 1. someone else's profile - remove project headers on profile with no projects  
 - [x] 2. user's profile - remove headers and 2nd cta on projects tab for a user with no projects
 - [x] 3. user's profile - remove header on empty FAVs tab (i'm not sure we need this header for favs tab at all even on someone else's profile, since it just repeats the tab above. you can keep it if it's simpler though)

### 1. A profile with no projects still has headers. 

<img width="626" alt="image" src="https://user-images.githubusercontent.com/53411/39717139-5805c046-5200-11e8-8d39-3873b6573df9.png">

### 2. two CTAs

<img width="902" alt="image" src="https://user-images.githubusercontent.com/53411/39717231-99875e62-5200-11e8-83bb-e86d78be17f9.png">

### 3. favs header

<img width="813" alt="image" src="https://user-images.githubusercontent.com/53411/39717442-45231752-5201-11e8-8148-bf8f98be5d90.png">

xmatthewx:
Thanks @mmmavis. How does this sound?... I will try and make a diagram to cover if/then scenarios. Ideally, I can find a few universal rules we can apply everywhere, so we don't have to worry about tracking lots of details.

mmmavis:
Yay. This is done!

PR #956, opened 2018-05-10
Profile page followup

mmmavis:
Related to #955 

I hope I covered everything...

### DIFF
--- a/components/profile-tab-group/profile-tab-group.jsx
+++ b/components/profile-tab-group/profile-tab-group.jsx
@@ -8,6 +8,7 @@ import HintMessage from '../../components/hint-message/hint-message.jsx';
 import ProjectList from '../../components/project-list/project-list.jsx';
 
 const TABS = {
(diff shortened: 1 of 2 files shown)


components/profile-tab-group/profile-tab-group.jsx
```diff
@@ -8,6 +8,7 @@ import HintMessage from '../../components/hint-message/hint-message.jsx';
 import ProjectList from '../../components/project-list/project-list.jsx';
 
 const TABS = {
+  // tab name: [ `project type 1`, `project type 2`, ... ]
```
gvn: Should this be commented out? Just delete?

mmmavis: Ah sorry, bad commenting style lol. That line is meant to be a comment. I can move it to just before `const TABS = {` if that can look less confusing 😂

gvn: Oh I see now...yeah that looks like a commented out property of TABS.

PR #961, opened 2018-05-18
Dependency updates

cadecairos:
This PR updates a variety of dependencies that had vulnerable paths.


patjouk:
I tested it locally and got this warning:
```
WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/concepts/mode/
```

I continued to test the code and plugged it to a local version of the API. Everything seems to work: I can post, do favs, etc.

But when I try to deploy the review app, it fails and I get this error:
```
Error: webpack.optimize.UglifyJsPlugin has been removed, please use config.optimization.minimize instead.
```

Those warning and error are probably linked to upgrading to webpack 4.

package.json
```diff
@@ -60,45 +60,47 @@
...
+    "babel-preset-es2015": "^6.24.1",
+    "babel-preset-react": "^6.24.1",
     "browserslist": "^3.2.6",
     "classnames": "^2.2.5",
     "document-env-vars": "^2.1.0",
     "dotenv": "^4.0.0",
     "express": "^4.16.2",
     "font-awesome": "^4.7.0",
     "helmet": "^3.9.0",
+    "lodash": "^4.17.10",
```
Pomax: what's using lodash? O_o

package.json
```diff
@@ -60,45 +60,47 @@
...
-    "react-select": "^1.1.0",
-    "react-tag-autocomplete": "^5.5.0",
+    "react-select": "^1.2.1",
+    "react-tag-autocomplete": "^5.5.1",
     "shx": "^0.2.1",
-    "snyk": "^1.53.0",
+    "snyk": "^1.80.1",
     "url": "^0.11.0",
-    "webpack": "^2.7.0"
+    "webpack": "^4.8.3"
```
Pomax: This is an impossibly big version update. Let's roll this one back, and file "updating webpack" as a separate issue, because there are breaking changes.

package.json
cadecairos: this is more than likely a dependency of another dependency, and we're pinning a version that's got a security fix.

package.json
cadecairos: :+1:

Pomax:
it's a good idea to have @mmmavis check this too =)

.gitignore
```diff
@@ -5,6 +5,7 @@ node_modules/
 dist/
 npm-debug.log
 *.env
+.idea/
```
mmmavis: @cadecairos what's this `idea` directory about?

Pomax: IntelliJ editor file (an annoyingly auto-generated file that we don't want on the repo)

patjouk:
You can add it to your local gitignore: `git/info/exclude`

.gitignore
mmmavis: ah kk

PR #964, opened 2018-06-07
Set default value for a required prop 'entryCount'

mmmavis:
Just did a prod push and I started getting this error when visiting some profiles. I assumed every profile returned from API call would have this `entry_count` property... I guess I was wrong!

screen shot 2018-06-07 at 12 00 01 pm


mmmavis:
AppVeyor seems to have some issues unrelated to this PR that's why the test is failing

> Error downloading cache item: Unable to connect to the remote server

I'm gonna merge this PR now.

Issue #944, opened 2018-04-19
Switch to use Pulse API v2


mmmavis:
Other than updating the endpoints we should also make sure data returned from v2 endpoints is what we expected to get.

Some endpoints to pay attention to:

- https://github.com/mozilla/network-pulse-api#get-apipulseprofilesidnumber-with-optional-formatjson
- (more to be added)

mmmavis:
Done!

PR #965, opened 2018-06-07
Use Pulse API v2

mmmavis:
Related to #944 

@alanmoo I couldn't decide if we should do this by updating the env var or updating the strings in https://github.com/mozilla/network-pulse/blob/master/js/service.js#L177-L272 so they have `/v2` in the API endpoints. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Pulse reveals the network's footprint, it fosters collaboration and amplifies th
 
 ### Testing
 
-Interested in contributing to this project? A good place to start is by helping with [Quality Tests](https://github.com/mozilla/network-pulse/wiki/Quality-Tests). File a [new issue](https://github.com/mozilla/network-pulse/issues) to say hello and let us know that you're interested.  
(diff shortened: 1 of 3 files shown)


alanmoo:
My instinct is that because this is a change in behavior (theoretically, if not in practice with this specific API upgrade) this change should happen in code. That way when we need to go from v2 to v3, it forces us to confirm the upgrade in a commit rather than casually doing it with an env var.


@cadecairos do you have thoughts?

cadecairos:
I agree with @alanmoo, lets choose our API version in code

README.md
```diff
@@ -12,7 +12,7 @@ Pulse reveals the network's footprint, it fosters collaboration and amplifies th
 
 ### Testing
 
-Interested in contributing to this project? A good place to start is by helping with [Quality Tests](https://github.com/mozilla/network-pulse/wiki/Quality-Tests). File a [new issue](https://github.com/mozilla/network-pulse/issues) to say hello and let us know that you're interested.  
+Interested in contributing to this project? A good place to start is by helping with [Quality Tests](https://github.com/mozilla/network-pulse/wiki/Quality-Tests). File a [new issue](https://github.com/mozilla/network-pulse/issues) to say hello and let us know that you're interested.
```
mmmavis: removed extra whitespace

mmmavis:
Cool cool PR updated! @alanmoo 👀 ?

mmmavis:
@alanmoo got a chance to review this?

PR #971, opened 2018-06-26
Showed 'about' tab on profile only when there's long bio to show

### LINKED ISSUE #969: My profile shows "About" tab even if empty
STR:

1. Log in, visit https://www.mozillapulse.org/profile/me
2. Even if you don't have extended profile info enabled, you'll see the "About" link and be at that route.
(For example, https://www.mozillapulse.org/profile/93 doesn't have it, unless I'm logged in)

Expected behavior:
If extended profile info isn't enabled, you shouldn't see the tab/be on that route.

cc @xmatthewx

mmmavis:
Fixed #969

### DIFF
--- a/components/profile-tab-group/profile-tab-group.jsx
+++ b/components/profile-tab-group/profile-tab-group.jsx
@@ -50,17 +50,19 @@ class ProfileTabGroup extends React.Component {
   }
 
   getAvailableTabs(props) {
-    // 2 scenarios
-    // [1] viewing your own profile: show all tabs
-    // [2] viewing other people's profile: show tab only if there is content to show
-    const TAB_NAMES = Object.keys(PROJECT_TYPES_BY_TAB_NAME);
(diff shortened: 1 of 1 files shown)


mmmavis:
@xmatthewx can you help with design review? The review app links to staging API so you can just go on to the staging Django dashboard and make your profile changes there for QA purposes.

alanmoo:
The code makes sense, though I don't think it can be tested currently in the review app- I'm having trouble logging in. Does Google auth need updating for a review app/staging to work or am I missing something?

mmmavis:
@alanmoo Login doesn't seem to work directly on review apps. I think the trick (at least this is how the way I've been using) is to log in/out from **[Pulse staging via front end](https://network-pulse-staging.herokuapp.com)**.

mmmavis:
@alanmoo R 👀 when you get a chance?

Issue #920, opened 2018-02-13
Build extend profiles (frontend edit form)


alanmoo:
to enable fellows and staff to edit their own extended profile info

codexdelta:
Hi, I would be interested in taking up this task.

alanmoo:
This need a bit more coordination for it to be ready to work on. In addition, it's on our priority roadmap, so one of the staff engineers will be taking it on. Please check out issues labeled "Help wanted" though!

@xmatthewx Do designs exist for this already?

mmmavis:
@xmatthewx would this be the next Pulse ticket i should work on?

xmatthewx:
I think yes. We should take a look at any comps @kristinashu did and draft a requirements list before you jump in. A few things that jump to mind: 

 - [ ] only show extended fields to people who have their extended profile flag checked
 - [ ] don't show these fields on frontend: profile type, program type, year. those should be maintained by staff via the admin, not by users.

mmmavis:
@xmatthewx

FYI, afaik, data returned from API doesn't directly tell you if a profile has `extended profile` toggled on or not. When `extended profile` of a profile is enabled and any meta in the "extended profile" section is available, API will automatically emit that info (e.g., include a `user_bio_long` property in the JSON).

xmatthewx:
Well, maybe that's okay. Unless I'm mistaken, I think long bio is the only field we would want to add to the form. We could just add that as a clearly optional field. And... that begs the question: do we need a flag to enable/disable extended profiles? If the "official" fields are only accessible via django, the flag might be unnecessary, even for use cases on foundation site. 

I'm curious what @mmmavis and @gideonthomas think. This is not a technical question, but more about how we expect users, admins, and sites to use pulse.

mmmavis:
(attaching a screencap here in case it's useful as a reference)
<img width="477" alt="screen shot 2018-06-26 at 10 39 46 am" src="https://user-images.githubusercontent.com/2896608/41929554-4bc138f2-792d-11e8-8783-be33bc309c32.png">

If the `user long bio` is the only field in that section we are allowing users themselves to change, then I agree with @xmatthewx that we could probably just get rid of the `extended profile` flag and make sure `profile type`, `program type`, `program year`, and `affiliation` fields are editable only via the Django dashboard.

[...]

PR #974, opened 2018-06-29
Added 'long bio' field to profile edit form

mmmavis:
Related to #920 

https://github.com/mozilla/network-pulse/pull/974/files?w=1

### DIFF
--- a/pages/profile-edit/form/fields.js
+++ b/pages/profile-edit/form/fields.js
@@ -7,80 +7,100 @@ let IssuesLabel = () => {
   return <div>Your <Link to="/issues" target="_blank">Key Internet Issues</Link> interest.</div>;
 };
 
-module.exports = {
-  "custom_name": {
-    type: `text`,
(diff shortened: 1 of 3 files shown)


mmmavis:
@xmatthewx 👀 ?

I had to relocate the hint message to below the field title due to technical/design restriction (there's no room below the textarea to fit it). Feel free to suggest better copy for it as I probably have grammar errors there 😂 

screen shot 2018-06-29 at 10 30 37 am
screen shot 2018-06-29 at 10 31 20 am

mmmavis:
@xmatthewx i'll remove hint but let's keep the character count until backend increases the character limit from `4096` to a more reasonable number. Otherwise users won't be able to post if their input long bio is longer than the limit (server error).

xmatthewx:
That copy looks good @gideonthomas. Could it be shortened a little? Either way, go for it.

mmmavis:
🎉

Issue #977, opened 2018-07-03
Newlines not being rendered in Profile About tab


mmmavis:
On staging:

![screenshot_2018-07-03 update profile - mozilla network pulse](https://user-images.githubusercontent.com/2896608/42242379-835ea908-7ec2-11e8-91a0-5ce6295901c0.png)

![screenshot_2018-07-03 mmm - mozilla network pulse](https://user-images.githubusercontent.com/2896608/42242374-8130a528-7ec2-11e8-86b8-7f6504c60137.png)

xmatthewx:
Interesting. It seems to work for paragraphs (2 newlines) and bullets (single newline), but not a simple single line break. 

It'd be good to fix this, but it's definitely not an emergency.

mmmavis:
@xmatthewx PR is in review already 👍

gideonthomas:
This can be closed?

mmmavis:
Fixed!

PR #978, opened 2018-07-03
Made sure newlines are rendered in Profile About tab

mmmavis:
Related to #977 

screenshot_2018-07-03 update profile - mozilla network pulse 1
screenshot_2018-07-03 localhost mavis ou - mozilla network pulse

### DIFF
--- a/components/profile-about-tab/profile-about-tab.jsx
+++ b/components/profile-about-tab/profile-about-tab.jsx
@@ -7,16 +7,23 @@ class ProfileAboutTab extends React.Component {
     super(props);
   }
 
+  renderLongBio() {
+    if (!this.props.longBio) return null;
+
+    return this.props.longBio.split(`\n`).map(line => {
+      return <ReactMarkdown
+        key={line}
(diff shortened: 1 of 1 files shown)


components/profile-about-tab/profile-about-tab.jsx
```diff
@@ -7,16 +7,23 @@ class ProfileAboutTab extends React.Component {
     super(props);
   }
 
+  renderLongBio() {
+    if (!this.props.longBio) return null;
+
+    return this.props.longBio.split(`\n`).map(line => {
```
gideonthomas: I guess we can't just replace the `\n` with `<br>` because we're escaping html?

mmmavis: Yup and `<br>` isn't a markdown syntax either.

mmmavis:
> I feel like this should've been supported in the ReactMarkdown component itself.

Yea that was my initial bet but I couldn't find any relevant documentation there.

PR #989, opened 2018-11-02
Redirect to pulse API admin

cadecairos:
Related issue: https://github.com/MozillaFoundation/mofo-devops/issues/599
Summary: Adds an Express route that redirects to the Pulse API admin.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,3 +1,4 @@
+import url from 'url';
 import express from 'express';
 import path from 'path';
 import helmet from 'helmet';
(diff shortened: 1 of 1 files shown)


patjouk:
I was mentioning those pages when you get redirect out of a domain to go to another one: "You're leaving X website, clear here to continue."

It's just that if you use the prod API as backend for local dev, you might do something bad by accident. The default is set up to pulse staging, but you never know.

cadecairos:
I'm not entirely sold on the need for an entire page and client-side redirect. But how about we only add the route if `NODE_ENV === 'production`

patjouk:
Yeah, that's a better and faster solution.

cadecairos:
@patjouk updated and ready for re-review

Issue #1011, opened 2019-02-21
Support query parameters for pre-filled tags


alanmoo:
In the interest of supporting Pulse for the organization of events, we should have the ability to direct users to an "add submission" URL in which one or more tags are automatically applied to the submission.

Suggested approach: `https://mozillapulse.org/add?tag=foo,bar,baz` results in a page where the tags `foo`,`bar`, and `baz` are pre-filled in the tags field.

"Probably a good idea to have" features:
* Only auto-fill the tags if they already exist in the tag database
* Tags pre-filled due to query string are unable to be removed from the "tag selection" field, but the submitter can add more. (Continued ability to add more tags is more important than preventing removal of pre-filled ones)

Building this will help meet the needs of this year's sprint/internet health month.

cc @acabunoc @zee-moz @xmatthewx @gideonthomas

gideonthomas:
@alanmoo to determine if this a "nice to have" or a core requirement:

People who are not logged in, when they hit that url should be redirected to the pulse login page and once they login, they should be redirected to the `/add?tags=...` page.
Setting that as the redirect path should be easy (requires a backend env var change). The only small caveat is if they need to verify their email with Github, they will not be redirected to the `add?tags=...` page ([filed here](https://github.com/mozilla/network-pulse-api/issues/416)). But I think this can be easily addressed with copy in the CTA.

alanmoo:
Great point. This is probably more important than the other "good to have" features.

alanmoo:
@gideonthomas It looks like this already works when logging in using Google and Github. So the only thing I haven't tested is email verification with Github- is that because it comes in as a different request?

gideonthomas:
@alanmoo the only thing that won't work is redirecting to the page you were on. Currently it will redirect you to the pulse homepage. Re: the github verification thing - I'm not sure why, but the simple answer is django-allauth has not implemented it.

alanmoo:
I still don't understand. This works:

![login-forwarding-pulse](https://user-images.githubusercontent.com/1682681/53200325-4570ff80-35ef-11e9-8625-117af6b59919.gif)

gideonthomas:
oh interesting...ignore me then :P

Pomax:
PR up on https://github.com/mozilla/network-pulse/pull/1015, I didn't see an easy to way to filter out "non-existent tags" and so it didn't seem worth trying to figure out.

xmatthewx:
[...]

PR #1015, opened 2019-02-22
added query parsing and tag list detection for the tag input field

Pomax:
Related issue: https://github.com/mozilla/network-pulse/issues/1011
Summary: adds support for presupplied tags using `add/?tags=a,b,c,etc`

### DIFF
--- a/js/utility.js
+++ b/js/utility.js
@@ -26,22 +26,6 @@ const Utils = {
     return console.error(`No window available to get the current url!`);
   },
 
-  getCurrentURLQuery() {
-    if (typeof window !== `undefined`) {
-      let map = {};
-      window.location.search.replace('?','').split('&').forEach(pair => {
-        let keyValue = pair.split('=');
-        if (keyValue[1]) {
-          let key = decodeURIComponent(keyValue[0]);
-          let value = decodeURIComponent(keyValue[1]);
-          map[key]=value;
-        }
-      });
-      return map;
-    }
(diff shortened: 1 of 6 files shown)


js/utility.js
```diff
@@ -26,6 +26,22 @@ const Utils = {
     return console.error(`No window available to get the current url!`);
   },
 
+  getCurrentURLQuery() {
```
gideonthomas: any reason we don't just use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams?

pages/add/form/auto-complete-input.js
```diff
@@ -30,6 +31,24 @@ export default class AutoCompleteInput extends Component {
     };
   }
 
+  /**
```
gideonthomas: probably needs to be in this file instead: https://github.com/mozilla/network-pulse/blob/master/pages/add/form/tags.js#L4

js/utility.js
Pomax: mostly that when last I checked, that wasn't broadly supported, and I never bothered to check if that had changed in the mean time! Good call =)

Pomax:
tested on https://network-pulse-staging-pr-1015.herokuapp.com/entry/462?tags=prefilled%2Ctags by hitting up https://network-pulse-staging-pr-1015.herokuapp.com/add?tags=prefilled,tags and then saving with only a title and url, no editing of the tags beyond what the autoloading does.

Issue #1040, opened 2019-04-26
Update tagline styling


natalieworth:
Update the "Discover & ..." tagline area to H2 and the 'Learn More' link:

<img width="825" alt="Screen Shot 2019-04-26 at 12 35 38 AM" src="https://user-images.githubusercontent.com/14876567/56791200-52f76100-67bb-11e9-82d9-54dcbc1b9305.png">

youriwims:
Closed by #1048

Issue #1036, opened 2019-04-25
Update site footer


natalieworth:
Update/ apply the foundation.mozilla.org footer to the Pulse site.

<img width="741" alt="Screen Shot 2019-04-24 at 7 37 02 PM" src="https://user-images.githubusercontent.com/14876567/56706058-5fe75800-66c8-11e9-8112-57ccd6f469b8.png">
<img width="275" alt="Screen Shot 2019-04-24 at 7 37 14 PM" src="https://user-images.githubusercontent.com/14876567/56706062-6249b200-66c8-11e9-871b-34095fe0b554.png">

cc @xmatthewx @mmmavis @youriwims @alanmoo @gideonthomas

gideonthomas:
for awareness during implementation: https://github.com/mozilla/foundation.mozilla.org/issues/3060

xmatthewx:
- [ ] link to foundation.mozilla.org, not mozilla.org in the footer. Both the visible link and the logo.

mmmavis:
@xmatthewx @alanmoo can we land this in 2 stages? Pulse doesn't have a newsletter sign up component yet. To reduce implementation & PR review complexity I'm hoping we can 1) land all non-form elements in first 2) implement sign up component & add it to the footer.

alanmoo:
Makes sense to me.

youriwims:
@xmatthewx which link should we use for the donate link in the Pulse footer? https://github.com/mozilla/network-pulse/pull/1054#discussion_r283459053

alanmoo:
Likely this: https://donate.mozilla.org/?utm_source=mozillapulse.org&utm_medium=referral&utm_content=footer

youriwims:
Awesome, thanks!

PR #1048, opened 2019-05-01
Added styles & Updated Tagline

### LINKED ISSUE #1038: Style standardization
From the [redpen mockups](https://redpen.io/p/vqdf0d6fb857181a19) it seems like we borrowed quite a few of styles from [Foundation site](http://foundation.mozilla.org/style-guide/). Before updating individual component's styling on Pulse (e.g., nav, footer, buttons, links etc), we should have style standardization in place first.

@natalieworth can you give a complete list of what we should include on Pulse's "style guide"?

For example,

- Colours? (Do we copy the entire colour variable set from Foundation site?)
- Typography?
- Link style?
- Button style?
- anything else?

/cc @youriwims

youriwims:
Closes #1038, #1040

Design flag: The styling rules that we're using from the Foundation-site currently state that links within an `h2` element should inherit the heading's font colour. Here, I had to change those rules in order for the `h2`'s link to be `$dark-blue`(This impacts all heading elements btw). Let me know if that's 🆒 @natalieworth 

This pr would also close the style standardization (#1038) ticket since we've imported the bulk of type, variable, and mixin styles.  /cc @mmmavis


natalieworth:
> Closes #1038, #1040
> 
> Design flag: The styling rules that we're using from the Foundation-site currently state that links within an `h2` element should inherit the heading's font colour. Here, I had to change those rules in order for the `h2`'s link to be `$dark-blue`(This impacts all heading elements btw). Let me know if that's 🆒 @natalieworth
> 
> This pr would also close the style standardization (#1038) ticket since we've imported the bulk of type, variable, and mixin styles. Let me know if this pr takes care of style standardizatio or if more styles need to be added /cc @natalieworth @mmmavis

Hey, thanks @youriwims I'll update that in my overrides. I think it's something we didn't originally think about for the Foundation site design since we didn't have a use case for it. I think it needs to be `$dark-blue` on the Pulse site so it's immediately recognizable as a link.

natalieworth:
@youriwims That looks really great. The type is awesome! I think I may want to increase the spacing a bit around it—but is that something you'd rather do later/ we could adjust the spacing in a new ticket (as I feel the search and help area may affect it too and we'll probably have to adjust spacing for those)

youriwims:
Okay @mmmavis, makes complete sense! I'll make those updates 👍

components/footer/footer.jsx
```diff
@@ -29,7 +29,7 @@ const Footer = () => {
     {
       name: `mozilla`,
       link: `https://mozilla.org`,
-      description: (<p>Mozilla is a global non-profit dedicated to putting you in control of your online experience and shaping the future of the web for the public good. Visit us at <a href="https://mozilla.org">mozilla.org</a>.</p>),
+      description: (<p className="dark-theme body-small">Mozilla is a global non-profit dedicated to putting you in control of your online experience and shaping the future of the web for the public good. Visit us at <a href="https://mozilla.org">mozilla.org</a>.</p>),
```
youriwims: small fix until the [footer updates](https://github.com/mozilla/network-pulse/issues/1036) are done.

alanmoo: Good catch

scss/main.scss
```diff
@@ -14,9 +15,11 @@
 @import '../node_modules/bootstrap/scss/grid.scss';
 @import '../node_modules/react-select/scss/default.scss';
 
-// mixins
+// Foundation Styles
 
-@import 'mixin.scss';
```
alanmoo: The /foundation/mixins.scss file doesn't seem to contain everything that this file did. Does everything still work properly/can we get rid of the mixin.scss file?

youriwims: Everything still works properly, this mixin file is the original pulse mixin file. I just moved it up to line 2 ↗️ to separate it from mofo-bootstrap & foundation styles.

alanmoo: Wow, I totally missed that.

PR #1054, opened 2019-05-03
Updated footer styles

### LINKED ISSUE #1036: Update site footer

youriwims:
Closes #1036

This pr takes care of updating the overall footer styles to match the foundation site's current footer.  Adding the sign up component will be done separately.


components/footer/footer.jsx
```diff
@@ -1,42 +1,86 @@
...
+          </a>
+        </li>
+      </ul>
+    );
+  }
+  renderFooterLinks() {
+    return (
+      <ul className="link-list list-unstyled mb-0">
+        <li className="col-auto mb-2" key="Donate">
+          <a className="footer-link-donate" href="https://donate.mozilla.org/?utm_source=foundation.mozilla.org&amp;utm_medium=referral&amp;utm_content=footer">Donate</a>
```
mmmavis: @xmatthewx any idea what utm code etc we should use for the Donate line on Pulse footer?

youriwims:
Sure,that works—thank you!

components/footer/footer.jsx
```diff
@@ -1,42 +1,86 @@
...
+          <a
+            className="github small d-inline-block"
+            href="https://www.github.com/mozilla/foundation.mozilla.org/">
+            <span className="sr-only">Github</span>
+          </a>
+        </li>
+      </ul>
+    );
+  }
+  renderFooterLinks() {
```
mmmavis: What are the classnames `footer-link-*` for? If they are not being used elsewhere let's remove these classnames. Don't forget we want these links to open in a new tab!

I would simplify repetitive code in this method into something like

```jsx

import React from "react";

// this also makes updating links in the future easier
const FOOTER_LINKS = [
    { text: `Donate`, link: `link to donate` },
    { text: `Legal`, link: `link to legal` }
    ...
    ...
];

export default class MofoFooter extends React.Component {
  ...
  renderFooterLinks() {
    let links = FOOTER_LINKS.map(link => {
        return <li className="col-auto mb-2" key={link.text}>
          <a href={link.link} target="_blank">{link.text}</a>
        </li>;
    });

    return (
        <ul className="link-list list-unstyled mb-0">
          {links}
        </ul>
      );
    }
  }
  ...
}
```

mmmavis: let's simplify `renderSocialLinks()` as well

components/footer/footer.jsx
```diff
@@ -1,42 +1,86 @@
...
-    }
-  ];
-
-  return (
-    <MofoFooter footerLinks={FOOTER_LINKS} orgs={ORGS} />
-  );
-};
-
-export default Footer;
+export default class MofoFooter extends React.Component {
```
mmmavis: let's change `MofoFooter` to `Footer` so it matches the file name

main.jsx
```diff
@@ -21,7 +21,7 @@ import ProfileEdit from './pages/profile-edit/profile-edit.jsx';
 import NotFound from './pages/not-found.jsx';
 
 import Navbar from './components/navbar/navbar.jsx';
-import Footer from './components/footer/footer.jsx';
+import MofoFooter from './components/footer/footer.jsx';
```
mmmavis: Let's just keep the same name `Footer`

components/footer/footer.jsx
mmmavis: @youriwims the link included here is specifically for the footer Donate link on Foundation site. Can you update the link after @xmatthewx replies?

components/footer/footer.jsx
```diff
@@ -1,42 +1,86 @@
...
+  renderSocialLinks() {
```
mmmavis: Let's simplify `renderSocialLinks()` as well. See https://github.com/mozilla/network-pulse/pull/1054/files#r283571169

components/footer/footer.scss
```diff
@@ -4,3 +4,16 @@ footer {
   bottom: 0;
   left: 0;
 }
+
+.site-footer {
+  background-color: $black;
+
+  .link-list {
+    @media (min-width: $bp-md) {
+      -webkit-column-count: 2;
```
mmmavis: We have `autoprefixer` npm task set up so we don't have to add vendor prefix ourselves... unless... that task is no longer working? 🙄

scss/main.scss
```diff
@@ -39,6 +45,10 @@ h2 {
   font-weight: 500;
 }
 
+hr {
+  border-color: #ccc !important;
```
mmmavis: Please really avoid using `!important`! 😬 We can always create new rule targeting a more specific selector.

scss/overrides/_type.scss
```diff
@@ -0,0 +1,18 @@
+a {
+  text-decoration: none;
+}
+
+.learn-more {
+  color: $dark-blue !important;
```
mmmavis: please don't use `!important`

components/footer/footer.jsx
youriwims: sounds good 👍

components/footer/footer.scss
youriwims: nope, sorry force of habit—I'll update that

scss/overrides/_type.scss
youriwims: For this, it's a one off and is inheriting the color of the h2 element. What would be a better way to target this specific element?

youriwims: btw this class is for the Pulse tagline

scss/main.scss
youriwims: 😞 sorry, will remember that!

components/footer/footer.jsx
alanmoo: Since this is a static component, is there really a benefit to having this be dynamically generated? In trying to initially migrate this to Pulse with Youri, this felt overly complicated and I suggested simply hardcoding the footer, since it's not being pulled in as part of the MoFo UI library anymore.

components/footer/footer.jsx
youriwims: Got it ➡️ updated link ➡️ https://github.com/mozilla/network-pulse/issues/1036#issuecomment-492395064
Code update: https://github.com/mozilla/network-pulse/pull/1054/files#diff-aed0ade64de8902416d11ddfa47d85d0R35

components/footer/footer.jsx
mmmavis: I thought about that too. But if we decide to hardcode this then later when we have to plug in the sign up form (which will be a React component) then we'll have to go with the Foundation site way that is using vanilla JS to target the node and ReactDOM to inject the form which I think might be less straightforward.

alanmoo: I'm not sure I follow. We need to trigger the React component somehow, what's the difference between triggering it as it is now or limiting the scope to just the form?

Though my suggestion was that we put the markup for the footer right into the render method, so it's still a React component but without the need to iterate over items and generate the HTML.

mmmavis: Ah. I thought you meant putting the static markup as HTML, in a `.html` file and use Foundation site's way to find a DOM node to inject the form portion of it.

If the code is all gonna be in one React component I _think_ putting the "static content" right into the render method vs create helper render method to iterate over links essentially are the same (as it's all JSX we are talking about here). I prefer the latter since it makes the code easier to maintain and increase code legibility (e.g., when we have to update a classname we won't have to update 6 instances of the same wrapper). With the array approach it does add more build time. But it's probably just a small fraction of ms we are talking about here and it's fast enough that we can simply ignore.

I don't have a strong opinion here as footer section is unlikely to be changed often. But I'm hoping we can utilize JS methods more as we are building React components. 😎

alanmoo: Yeah, I'm thinking more about engineer time to make the logic work vs the time to copy/paste a piece of markup. Sounds like Youri should go with whichever approach she prefers/is easier?

youriwims: @alanmoo @mmmavis if you guys are ok with it, I'm going to keep the logic as is if there aren't any syntax/logic errors please give this a +1 when you have a chance.

mmmavis:
@youriwims is this ready for review again? If so can you `update branch`? I was gonna press that button but wasn't sure how your branch was set up and I didn't wanna accidentally create merge conflicts.

components/footer/footer.scss
```diff
@@ -4,3 +4,18 @@ footer {
...
+
+  .link-list {
+    @media (min-width: $bp-md) {
+      column-count: 2;
+      break-inside: avoid-column;
+    }
+  }
+
+  hr {
+    border-color: #ccc;
```
mmmavis: Please only use color variables for any colours (in this case`$gray-20`). I'm gonna add stylelint & prettier to this repo soon so things are more aligned with the rules we set for Foundation site.

scss/overrides/_mixins.scss
```diff
@@ -0,0 +1,7 @@
+@mixin body-text($font-family, $font-weight, $color, $dark-theme-color) {
```
mmmavis: `body-text` mixin in `scss/foundation-styles/mixins.scss` should already have this covered?

scss/overrides/_type.scss
```diff
@@ -0,0 +1,18 @@
+a {
```
mmmavis: If we are loading our foundation scss correctly, we shouldn't need these overrides for `a` here. Can you double check?

scss/overrides/_mixins.scss
youriwims: yep, you're right. 🌬 deleted

scss/overrides/_type.scss
youriwims: Right, but I think we talked about this somewhere and can't find it 😅 I've tried to find out where this is coming from and from what I can see, it's not coming from any styles in this codebase, mofo-ui, or mofobootstrap. It's highhhh up on line 70 of our main.scss with other sitewide rules. Do you have an idea of where it may be coming from? 
Screen Shot 2019-05-17 at 2 08 17 PM

youriwims: p.s. the foundation-site doesn't have this issue because it doesn't have the same inital code in the compiled css.

mmmavis: It's risky to add override for all `a` in this PR. Let's remove this entire chunk from this file and add the same rules to `footer.scss` targeting `.site-footer a {}` for now. I'll file a new ticket to track down where that `text-decoration: underline` rule for `a` is coming from.

mmmavis: ticket filed: https://github.com/mozilla/network-pulse/issues/1072

scss/foundation-styles/mixins.scss
```diff
@@ -40,6 +40,8 @@
 
     a {
       @include dark-theme-inline-anchor($light-blue);
+
+      text-decoration: none;
```
mmmavis: we shouldn't need this?

scss/overrides/_type.scss
youriwims: this rule is coming from `dist/css/main.css`...as far as if it's being compiled from elsewhere, not sure...but that's a starting point i guess.

scss/foundation-styles/mixins.scss
mmmavis: we shouldn't need this?

mmmavis:
@natalieworth tagging you again to see if we need to make any temporary styling adjustment in the footer in the absence of the to-be-implemented sign up form.

https://network-pulse-staging-pr-1054.herokuapp.com

(form work will be tackled as follow-up)

youriwims:
@mmmavis I already re-requested you but feel free to wait till @natalieworth is done with hers. 🙏thankyouthankyou

scss/overrides/_type.scss
mmmavis: I'll update that ticket with the findings I discovered!

youriwims:
Thank you both @mmmavis & @natalieworth  for your patience 🙏  ➡️ Merging nowwwww.

Issue #1042, opened 2019-04-26
Add a help dropdown


natalieworth:
The 'help' drop down will include all of the topics that people are able to volunteer/ help with in Pulse (note that I've included 'help' as an additional/ the original (default) for people to view everything):
Help
Attend
Create content
Code
Design
Fundraise
Join community
Localize & translate
Mentor
Plan & organize
Promote
Take Action
Test & feedback
Write documentation

natalieworth:
<img width="699" alt="Screen Shot 2019-04-26 at 12 44 11 AM" src="https://user-images.githubusercontent.com/14876567/56791735-a6b67a00-67bc-11e9-8145-fd963f80f7af.png">
<img width="698" alt="Screen Shot 2019-04-26 at 12 44 18 AM" src="https://user-images.githubusercontent.com/14876567/56791738-a8803d80-67bc-11e9-8b58-6d593b475d80.png">
<img width="698" alt="Screen Shot 2019-04-26 at 12 44 32 AM" src="https://user-images.githubusercontent.com/14876567/56791741-aae29780-67bc-11e9-8972-f38989942a34.png">

natalieworth:
<img width="260" alt="Screen Shot 2019-04-26 at 12 43 38 AM" src="https://user-images.githubusercontent.com/14876567/56791752-b46bff80-67bc-11e9-9c7c-f7d0e11e87b0.png">
<img width="260" alt="Screen Shot 2019-04-26 at 12 43 46 AM" src="https://user-images.githubusercontent.com/14876567/56791755-b59d2c80-67bc-11e9-8635-8745b5c7d7fb.png">
<img width="261" alt="Screen Shot 2019-04-26 at 12 43 53 AM" src="https://user-images.githubusercontent.com/14876567/56791764-b930b380-67bc-11e9-8461-47e0e70b8789.png">

alanmoo:
We should pull the list from `{api}/api/pulse/v2/helptypes/` to be consistent.

xmatthewx:
On this main list view, a user can search, pick a help cat, or tap a trending tag. All of these actions should clear each other, or not, depending. For example, if a user is viewing a search of foo, and then they tap a tending tag, search should clear. When a user selects a Help category, Nat and I are fine if it clears the search or further filters it, as long as UI updated appropriately. Combined filtering would be great, BUT only if it's trivial to build. I prefer the simplest path to shipping this. We can always enhance later.

alanmoo:
[...]

PR #1070, opened 2019-05-17
Adding Help Dropdown

youriwims:
Closes: #1042

View dropdown here: https://network-pulse-staging-pr-1070.herokuapp.com/search


natalieworth:
@youriwims @alanmoo For the select element, what browsers allow for styling of it? Is it possible to make it so it's not the default?

This looks great!:
Screen Shot 2019-05-28 at 3 55 16 PM

Wondering if it's possible dev-wise to update this with some more styling? How hard is that to do/ what browsers are able to support it?
Screen Shot 2019-05-28 at 3 53 06 PM

alanmoo:
Virtually none; the open state is handled by the browser. In a very few cases you can change the color of the text and the background of it, but that’s it.

natalieworth:
> Virtually none; the open state is handled by the browser. In a very few cases you can change the color of the text and the background of it, but that’s it.

Okay, thanks Alan. @xmatthewx are we okay to leave this styling like this for now, but update in future revision? I don't love the default styling, but wondering what is a priority here?

xmatthewx:
Let's go with the browser defaults for now. We can see if this feature gets lots of action before we override it.

mmmavis:
For reference if we really want to build a pretty/custom dropdown we can rely on external library like [`react-select`](https://github.com/JedWatson/react-select) to build something that works like a dropdown but it's not semantically a standard browser dropdown.

e.g., Moderation view's dropdowns are built using that library.

react-select

alanmoo:
Yeah and they don’t quite work right on mobile 🙃

mmmavis:
@xmatthewx what's the expected behaviour for the help drowndown on people search tab?

alanmoo:
Am I the only one seeing this? 
image

components/help-dropdown/help-dropdown.scss
```diff
@@ -0,0 +1,18 @@
+#help-dropdown {
```
alanmoo: Let's not target using IDs; it makes overriding much harder later on if necessary.

pages/search/search.jsx
```diff
@@ -89,6 +108,17 @@ class Search extends React.Component {
     </div>;
   }
 
+  renderHelpFilter() {
+    let renderedHelpTypes = this.state.helpOptions.map(type => {
+      return <option key={type} value={type}>{type}</option>;
+    });
+
+    return <select id="help-dropdown" className="body-large p-3" value={ this.state.helpType } onChange={  (event) => this.handleHelpChange(event) }>
```
alanmoo: Mixing Bootstrap classes and a custom set of targeted styles makes for confusing debugging. Let's stick to targeted styles for this element (but generic as a class- we could re-use this style for other select elements, right?)

pages/search/search.jsx
```diff
@@ -4,11 +4,15 @@ import qs from "qs";
 import DebounceInput from 'react-debounce-input';
 import ReactGA from 'react-ga';
 import SearchTabGroup from '../../components/search-tab-group/search-tab-group.jsx';
+import Service from '../../js/service';
 
 class Search extends React.Component {
```
mmmavis: I would remove dropdown related code from this file and make it a component (ie. the existing`components/help-dropdown/help-dropdown.jsx` though this file is currently not being used anywhere). State like `helpOptions` is irreverent to the Search page component (Search page only needs to know the selected help type) and value updates to `this.state.helpOptions` will cause the Search page to re-render which isn't something we want.

mmmavis:
@xmatthewx I browsed the mockups for the new homepage on Redpen again but I'm still confused about how the help dropdown works. As far as I know, help type is associated to projects (and not profiles). If that's the case, should we make the help dropdown visible only when user's on the "Projects" search tab?

pages/search/search.jsx
mmmavis: Selection in the HelpDropdown component will have to trigger Search page's state update. You can use this [article](https://ourcodeworld.com/articles/read/409/how-to-update-parent-state-from-child-component-in-react) as a reference to get you started.

pages/search/search.jsx
youriwims: yep, you're right 💯

components/help-dropdown/help-dropdown.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+export default class HelpDropdown extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+       options: [],
+       value: ''
+     };
+
+    this.handleChange = this.handleChange.bind(this);
```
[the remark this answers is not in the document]
youriwims: 🆗 so I don't have to bind my method if I do the arrow function in `onChange`? 🆒 🆒

components/help-dropdown/help-dropdown.jsx
```diff
@@ -0,0 +1,44 @@
...
+    // when user selects, value is updated
+    this.setState({ 
+      value: e.target.value
+    });
+
+    //value stored and used in path
+    let value = this.state.value.toLowerCase().replace(/\s/g,'');
+    let pathUpdate = `/?helptype=${ value }`;
+
+    //Update URL
```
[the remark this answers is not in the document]
youriwims: ah yes.

pages/search/search.jsx
youriwims: ah ok 🙃 thanks for the article!

xmatthewx:
> should we make the help dropdown visible only when user's on the "Projects" search tab?

@mmmavis did you chat with Nat? All clear now?

components/project-loader/project-loader.jsx
```diff
@@ -142,7 +142,7 @@ class ProjectLoader extends React.Component {
 
   renderEntryCounter() {
     return <SearchResultCounter
-      searchKeyword={this.props.search}
+      searchKeyword={this.props.search ? this.props.search : this.props.help_type}
```
mmmavis: Let's pass `help_type` as a new prop.

components/search-result-counter.jsx
```diff
@@ -5,10 +5,14 @@ export default (props) => {
```
mmmavis: I double checked with Nat - the pill button is only for displaying "help type". For text(keyword) search we can stick with existing design.

pages/search/search.jsx
```diff
@@ -48,6 +52,24 @@ class Search extends React.Component {
...
+  handleHelpChange(event) {
+    let helpType = event.target.value;
+
+    ReactGA.event({
+      category: `Search`,
+      action: `Help filtered`,
+      label: helpType
+    });
+
+    this.setState(
```
mmmavis: Let's add a helper method so both `handleHelpChange` and `helpInputChange` can utilize

something like:
```jsx
  setSearchCriteria(key, value) {
    let criteria = {
      keywordSearched: ``,
      helpType: ``
    };

    criteria[key] = value;

    this.setState(criteria, () => {
      this.updateBrowserHistory();
    });
  }
```

pages/search/search.jsx
```diff
@@ -38,8 +40,10 @@ class Search extends React.Component {
   updateBrowserHistory() {
     let keywordSearched = this.state.keywordSearched;
     let location = { pathname: this.props.location.pathname };
-    let query = {};
-
+    let query = {
```
mmmavis: Let's add `helpType` to `query` only if `helpType` exists

components/help-dropdown/help-dropdown.jsx
```diff
@@ -0,0 +1,31 @@
...
+    });
+  }
+
+  render() {
+    let renderedHelpTypes = this.state.helpOptions.map(type => {
+      return <option key={type} value={type}>{type}</option>;
+    });
+
+    return (
+      <select className="body-large dropdown" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
```
mmmavis: Let's rename `dropdown` to `help-dropdown` so it matches the component name. If in the future we decide to reuse the styling for other dropdowns on the site we can also make code adjustment accordingly.

pages/search/search.jsx
youriwims: Similar to Line 47? ⬇️

mmmavis: yup yup

pages/search/search.jsx
youriwims: what should the purpose be of the helper method?

mmmavis: So instead of having 

(in `handleHelpChange`)
```jsx
    this.setState(
      {
        helpType: helpType, 
        keywordSearched: `` 
      }, () => {
        this.updateBrowserHistory();
      });
```

and (in `helpInputChange`)
```jsx
    this.setState(
      {
        keywordSearched: keywordsEntered,
        helpType: ``
      }, () => {
        this.updateBrowserHistory();
      });
```

1. add a new helper method like what I suggested in my previous comment
2. replace the above code snippet with a function call to `this.setSearchCriteria`

e.g., ```this.setSearchCriteria(`helpType`, helpType)```

youriwims:
@natalieworth, there's been quite a few code updates + ux changes, so can you re-review this to make sure the dropdown still aligns with your designs? 🙏

components/search-result-counter.jsx
```diff
@@ -5,10 +5,23 @@ export default (props) => {
...
   let searchKeyword = props.searchKeyword;
+  let helpType = props.helpFilter;
+  let counterTextWQuery = `${counterText} for `;
 
   if (searchKeyword) {
-    counterText = `${counterText} for ${searchKeyword}`;
+    counterText = counterTextWQuery + searchKeyword;
   }
+  
+  if (helpType) {
```
youriwims: Is there a better way to write this?

natalieworth:
Screen Shot 2019-06-05 at 12 25 44 PM
I get the style updated a bit in firefox which is cool!

components/search-result-counter.jsx
alanmoo: Seems reasonable

components/help-dropdown/help-dropdown.scss
```diff
@@ -0,0 +1,11 @@
+.custom-select {
```
alanmoo: This is for custom select elements across the site, not just for the `help-dropdown` component. So let's put this somewhere/name it more appropriately?

components/help-dropdown/help-dropdown.scss
```diff
@@ -0,0 +1,11 @@
...
+  display: block;
+  width: 100%;
```
alanmoo: Given that this is going to be used everywhere, does 100% width make sense? Might be fine to leave now, but once we use this elsewhere it we'll probably need to reshape this a little bit.

components/help-dropdown/help-dropdown.scss
```diff
@@ -0,0 +1,11 @@
+.custom-select {
+  display: block;
```
alanmoo: Just out of curiosity: what's block doing for you that inline-block wasn't?

pages/search/search.scss
```diff
@@ -0,0 +1,19 @@
...
+    padding: 1px 8px;
+    text-transform: capitalize;
+    font-weight: 300;
+    margin-left: 4px;
+  }
+
+  .query-text {
+    @include scaleText( 12px, 16px, 16px, 24px);
+
+    display: flex;
```
alanmoo: What's the need for this?

pages/search/search.jsx
```diff
@@ -113,8 +153,9 @@ class Search extends React.Component {
         <Helmet><title>{this.state.keywordSearched}</title></Helmet>
         { this.renderSearchControls() }
         <SearchTabGroup
-          activeTab={this.state.activeTab}
+          activeTab={this.state.helpType ? `projects` : this.state.activeTab}
```
alanmoo: Nice

pages/search/search.jsx
```diff
@@ -29,25 +30,61 @@ class Search extends React.Component {
...
-      activeTab: props.match.params.tab
+      activeTab: props.match.params.tab,
+      helpType: query.helpType
     };
 
     return criteria;
   }
 
+  setSearchCriteria(key, value) {
+    let criteria = {
```
alanmoo:
```suggestion
	// Current UX dictates that either search or help filter can be used, but not both at the same time, so we reset things every time this function gets called
    let criteria = {
```

components/search-tab-group/search-tab.jsx
```diff
@@ -2,13 +2,13 @@ import React from 'react';
...
 export default (Loader, type) => (props) => {
-  const { keywordSearched } = props;
+  const { keywordSearched, helpType } = props;
   type = type.toLowerCase();
 
   return (
     <div className={`search-${type}-tab`}>
       <Helmet><title>{type.charAt(0).toUpperCase()+type.slice(1)} search</title></Helmet>
-      { keywordSearched && <Loader search={keywordSearched} showCounter={true} />}
+      <Loader search={keywordSearched} showCounter={true} help_type={helpType}/>
```
alanmoo: Why did the first part of this line go away? I can vaguely imagine why, but it's not entirely clear.

components/search-tab-group/search-tab-group.jsx
```diff
@@ -84,7 +88,7 @@ class SearchTabGroup extends React.Component {
     return SearchTab(
```
alanmoo: I know this was in place already, but what is going on here? The syntax on the line you changed is messing with my head.

components/help-dropdown/help-dropdown.scss
youriwims: removed that...literally it's doing nothing extra.

pages/search/search.scss
youriwims: This vertically centers the pill outline with the rest of the query-text and makes it more narrow to match design.

components/search-tab-group/search-tab.jsx
youriwims: I think it's because the  loader would only produce results if a keyword was searched e.g. if you selected something form the dropdown, no results would appear until you used the search bar. This change was made earlier when the dropdown and search worked together, but I think even with them working separately, this still may be necessary 🤔

components/search-tab-group/search-tab-group.jsx
alanmoo: Oh. `SearchTab` returns a function. That makes more sense.

components/search-tab-group/search-tab.jsx
alanmoo: Yup, this makes more sense now

pages/search/search.scss
alanmoo: Hm, ok. It feels slightly weird, but I can't put my finger on a reason so I'll let it go.

components/help-dropdown/help-dropdown.jsx
```diff
@@ -22,7 +23,7 @@ export default class HelpDropdown extends React.Component {
     });
 
     return (
-      <select className="custom-select" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
+      <select className="custom-select w-100" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
```
alanmoo: 👍

pages/search/search.scss
youriwims: Yeah, I know...but this is where that comes into play https://github.com/mozilla/network-pulse/pull/1070/files#diff-9b30595288e92d1997d202ae6f2ca348R17 Of course, if it was just text, there wouldn't really be a problem, but since it's text inside of a capsule I decided to do it this way primarily for aesthetic tbh ...it's a 2px diff 🤷‍♀

Issue #1074, opened 2019-05-24
Update icons on Pulse site with new iconography


natalieworth:
[extra-small-icons.zip](https://github.com/mozilla/network-pulse/files/3231020/extra-small-icons.zip)
### Final design + redpen link
Final design is here (and updated in the Sketch file):
https://redpen.io/p/wp17e9f94b044da4a0

On Pulse site:
- [x] Update 'down chevron arrow' on the 'help' drop down
- [ ] Update 'x' close/ cancel icon
- [x] Update the heart 'fav' icon
- [ ] Update search icon
- [ ] Add a profile icon beside the user name/ signup area

Desktop:
![02_Landing-page Copy@2x](https://user-images.githubusercontent.com/14876567/58532058-4474de80-8199-11e9-9136-65cccbe7897b.png)
![05_People-Search@2x](https://user-images.githubusercontent.com/14876567/58534812-1f846980-81a1-11e9-82c5-e848e0b85d5a.png)

Mobile:
![M  02_Landing-page Copy](https://user-images.githubusercontent.com/14876567/58532680-24461f00-819b-11e9-8a1f-a935b44851ac.png)
![M  Menu (signed out) Copy](https://user-images.githubusercontent.com/14876567/58532682-24deb580-819b-11e9-872d-ee4f4d5272b6.png)
![M  05_People-Search](https://user-images.githubusercontent.com/14876567/58534801-15fb0180-81a1-11e9-84e8-19d4fc8affcf.png)

------

### Original design ticket + related tickets
#2890 

------

### Assets
**Medium assets are here (chevron, close, heart, profile, and search icons)**
[medium-icons.zip](https://github.com/mozilla/network-pulse/files/3218878/Join.Page.Icons.zip)

**Extra small assets are here (linkedin, twitter, location icon)**
[extra-small-icons.zip](https://github.com/mozilla/network-pulse/files/3231023/extra-small-icons.zip)

----------

### Hover states
No hover on **chevron icon**
No hover on **'x' close/ cancel icon**
**Heart icons** are `$red` on save (no hover function)
Hover state of dark-blue #0d10bf for the **profile icon**
No hover on **search icon**

cc @youriwims @mmmavis @alanmoo @xmatthewx

mmmavis:
We should be able to clean up some SCSS overrides for the icons (especially around positioning) as the new icons are all in the same square size.

youriwims:
[...]

PR #1090, opened 2019-05-31
Bookmark heart icon style updates

youriwims:
Related issue: #1074 

**Design Deviation Summary** (_could be subject to brand update in future_): 
- Current dimensions are 24x24 vs design 19x22

@natalieworth just wanted to summarize ☝️ what we talked about


youriwims:
@natalieworth I noticed that some of the designs in Abstract are using the old heart icon. You may want to update them to the new one you've designed so that the iconography in the card design is consistent across all instances. 😸

youriwims:
@natalieworth ⬆️

PR #1094, opened 2019-06-05
Update anchor default styles

### LINKED ISSUE #1072: Clean up anchor <a> style so by default it doesn't have underline
Related to https://github.com/mozilla/network-pulse/pull/1054#discussion_r285205354

1. figure out where `a { text-decoration: underline; }` rule in the compiled css file is coming from
2. make sure by default anchors don't have underline
3. other other `text-decoration` related overrides in the codebase accordingly.

youriwims:
Closes  #1072 

_*Updated `@mixin link-text-decoration` to `text-decoration: none` default, but underline on hover_

### DIFF
--- a/scss/mixin.scss
+++ b/scss/mixin.scss
@@ -18,7 +18,7 @@
   }
 }
 
-@mixin link-text-decoration($normal-state: $link-decoration, $hover-state: $link-decoration) {
+@mixin link-text-decoration($normal-state: none, $hover-state: underline) {
   text-decoration: $normal-state;
 
   @include hover-focus-active {
--- a/scss/mixin.scss
+++ b/scss/mixin.scss
@@ -18,7 +18,7 @@
   }
 }
 
-@mixin link-text-decoration($normal-state: none, $hover-state: underline) {
+@mixin link-text-decoration($normal-state: none, $hover-state: $link-decoration) {
   text-decoration: $normal-state;
 
   @include hover-focus-active {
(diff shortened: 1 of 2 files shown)


scss/overrides/_variables.scss
```diff
@@ -21,7 +21,6 @@ $form-error-color: #d24735;
 $body-color: $black;
 $link-color: inherit;
 $link-hover-color: darken($mobile-blue, 15%);
-$link-decoration: underline;
```
mmmavis: let's keep this

mmmavis: let's keep this

scss/mixin.scss
```diff
@@ -18,7 +18,7 @@
   }
 }
 
-@mixin link-text-decoration($normal-state: $link-decoration, $hover-state: $link-decoration) {
+@mixin link-text-decoration($normal-state: none, $hover-state: underline) {
```
mmmavis: `@mixin link-text-decoration($normal-state: none, $hover-state: $link-decoration) { ... `

mmmavis:
@youriwims do you know if we want ALL anchors to 1) not have any `text-decoration` by default 2) and have underline only on hover? If so, we might be able to clean up the SCSS further.

youriwims:
@mmmavis if we're aligning with foundation styles, then yes. @natalieworth can you confirm, pls?

youriwims:
@natalieworth ⬆️

youriwims:
@mmmavis I'm going to file a ticket for enhancement to see if we can clean up `<a>` styles further so this can be landed. (will update this comment with ticket#)

Issue #1092, opened 2019-06-03
Join Us Page modal


natalieworth:
We will be creating a 'Join Us' page to onboard future (potential) fellows and mozilla community members with our email list and the Pulse site. Users will access the page through the ['join us' page](https://redpen.io/yg8a9572ea97403818) (on the [foundation/ cms site](https://foundation.mozilla.org/cms/pages/396/edit/)) where they will be taken to the ['sign-up' page ](https://redpen.io/whd4aa98717b56cd84)and then the modal walkthru on Pulse. 

Design has been approved and designs are updated in Abstract/ Sketch file.

---

**🗒  Design -> Dev Handoff Checklist**
(Feel free to remove items that are not applicable.)

- [x] Design has been finalized
- [x] Link to original design ticket
- [x] Link to related ticket
- [x] Link to Redpen
- [x] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included
- [x] Assets are included
- [x] Hover state for elements are designed
- [x] Other states for elements are designed (e.g., error state for form)

----

### Original design/ related tickets:
Design ticket for Join page: https://github.com/mozilla/foundation.mozilla.org/issues/2890
Join Us landing page implementation ticket on CMS: https://github.com/mozilla/foundation.mozilla.org/issues/3147

----

### Final designs are in [Redpen](https://redpen.io/p/wp17e9f94b044da4a0) ticket:
**Desktop:** https://redpen.io/whd4aa98717b56cd84
**Mobile:** https://redpen.io/vwc2a6d249220ef50f

----

### Assets:
**New icon of 'x/ close' icon for the modal:**
[close-icon.svg.zip](https://github.com/mozilla/network-pulse/files/3249758/close-icon.svg.zip)

**Profile image placeholder:**
[profile-image (blank).svg.zip](https://github.com/mozilla/network-pulse/files/3249796/profile-image.blank.svg.zip)

----

### Hover states
_Hover states follow the brand guidelines for the Foundation site button styles_
***The **exception** is the **['select image' button](https://redpen.io/ad71776d67900eef56)** has a hover state of  # B7B9FA

----

### Error states
**Modal form error states:**
Desktop: https://redpen.io/by1eb9d156ea6942e2
Mobile: https://redpen.io/tqf62683d4b9f88ce4

**Profile form error state:**
Desktop: https://redpen.io/axd7f6cad43d9d71cb
Mobile: https://redpen.io/mqfbef1094ad9785b3

----

[...]

PR #1097, opened 2019-06-11
Join Us Flow

mmmavis:
Related issue: #1092 

Form doesn't look same as what's in the mockups. For consistency I tried sticking with the same styling & form builder module that we used elsewhere on the site.

### Ideal Flow:

- visit a Pulse url with `sign_up=true` query param. e.g., https://network-pulse-staging-pr-1097.herokuapp.com/?sign_up=true
- after signing in (or creating an account) you should be auto redirected back to the Pulse page you were previously on (url should include query param `edit_new_profile=true` , e.g., https://network-pulse-staging-pr-1097.herokuapp.com/?edit_new_profile=true). Profile edit modal should pop up.

### Test Flow

Due to technical limitation with auth and a bug with local Pulse API not returning media url in absolute URL, this PR will need to be tested as the following

1. Check out this PR locally
2. Have local Pulse API and [social auth set up](https://github.com/mozilla/network-pulse-api#setting-up-social-authentication)
3. Use the following env vars in Pulse front-ends `.env`
```
PULSE_API=http://test.example.com:8000/api/pulse
PULSE_LOGIN_URL=http://test.example.com:8000/accounts/login/
PULSE_LOGOUT_URL=http://test.example.com:8000/accounts/logout/
```
4. Run Pulse API and Pulse locally
5. Go to http://test.example.com:3000/latest?sign_up=true
6. If you are not signed in, sign in with Google
image

7. You should now be redirected to http://test.example.com:3000/latest?edit_new_profile=true and modal should pop up.
8. Go through the profile edit flow in modal and see if there's any issue. (`Continue` button triggers form validation whereas `Skip` button does not)

Note: there's a bug on API's end about not sending media URL in absolute format, and hence the default thumbnail can't be displayed properly. See screencap below

image

### Check out how the modal and form work using review app

1. make sure you are logged into staging Pulse API
  - you can do it by going to the review app https://network-pulse-staging-pr-1097.herokuapp.com
  - use the sign in/up link at top right
2. visit any page on this review app with query param `edit_new_profile=true` (e.g., https://network-pulse-staging-pr-1097.herokuapp.com/featured?edit_new_profile=true)
3. play around with the modal and form (form won't submit due to staging API blocking review apps from POSTing)


mmmavis:
@Pomax @youriwims tagging you both here so you can lend a hand ✋ in case we need to polish this up for all-hands (I don't know what the plan is 😬 ). More details are in the PR description.  

I didn't wanna clean up my code further before design/UI/UX/content changes are finalized. Functionalities should all be working but code can for sure be further refactored depending on the to-be-finalized changes.

Pomax:
hm, I didn't get redirected. I started on https://network-pulse-staging-pr-1097.herokuapp.com/?sign_up=true, clicked google login, and then got sent to https://pulse-api.mofostaging.net after.

Pomax:
Hm, there are no "back" buttons during the initial flow, which feels like they're missing.

youriwims:
@Pomax Agreed.

components/join-us-modal/form/create-fields.js
```diff
@@ -0,0 +1,73 @@
...
+        validator.maxLengthValidator(70)
+      ]
+    },
+    email: {
+      type: `text`,
+      defaultValue: user.email,
+      fieldClassname: `form-control`
+    },
+    newsletter: {
+      // TODO:FIXME: not sure if Pulse API supports this?
```
Pomax: this is supported.

https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/urls.py#L73

components/join-us-modal/join-us-modal.jsx
```diff
@@ -0,0 +1,332 @@
...
+      () => {
+        if (currentStep === this.totalStep) {
+          this.handleFormSubmit();
+        }
+      }
+    );
+  }
+
+  renderProgressDots() {
+    let dots = [...Array(this.totalStep)].map((dot, i) => {
```
Pomax: this is still a beautiful piece of code.

components/join-us-modal/join-us-modal.jsx
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  renderSteps() {
+    return (
+      <div>
+        <Step
+          show={this.state.currentStep === 1}
+          heading="Complete your sign up"
+          subhead="Please confirm your information to finish signing up."
+          hint="(This information came from your Google or GitHub account)"
```
Pomax: can we know at this point whether this was github or google? Or do we only know "that this was a valid login"

components/join-us-modal/step.jsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+class Step extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  componentDidMount() {}
```
Pomax: these empty functions should not be necessary

components/navbar/navbar.scss
```diff
@@ -32,7 +32,7 @@
     width: 100%;
     height: 100%;
     background: rgba(0, 0, 0, 0.9);
-    z-index: 1000000;
+    z-index: map-get($z-indexes, nav-link-list);
```
Pomax: language nit: the plural of index is indices

package.json
```diff
@@ -26,7 +26,7 @@
     "optimize:svg": "find assets -type f -name '*.svg' -print0 | xargs -0 -n 1 -P 6 svgo --multipass --pretty --enable=removeTitle",
     "optimize:png": "find assets -type f -name '*.png' -print0 | xargs -0 -n 1 -P 6 optipng",
     "watch:css": "chokidar 'components/**/*.scss' 'pages/**/*.scss' 'scss/**/*.scss' -c 'npm run build-client:scss'",
-    "watch:js": "chokidar 'app.jsx' 'main.jsx' 'js/**/*.js' 'components/**/*.jsx' 'pages/**/*.jsx' 'pages/**/*.js' 'webpack.config.js' -c 'npm run build-client:js'",
+    "watch:js": "chokidar 'app.jsx' 'main.jsx' 'js/**/*.js' 'components/**/*.jsx' 'components/**/*.js' 'pages/**/*.jsx' 'pages/**/*.js' 'webpack.config.js' -c 'npm run build-client:js'",
```
Pomax: globs should understand `?` as "optional", so the following _should_ work?
```suggestion
    "watch:js": "chokidar 'app.jsx' 'main.jsx' 'js/**/*.js' 'components/**/*.jsx?' 'pages/**/*.jsx?  'webpack.config.js' -c 'npm run build-client:js'",
```

components/join-us-modal/join-us-modal.jsx
```diff
@@ -0,0 +1,332 @@
...
+Modal.setAppElement("#app");
+
+const SIGN_UP_PARAM = `sign_up`;
+const EDIT_NEW_PROFILE_PARAM = `edit_new_profile`;
+
+class JoinUsModal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.totalStep = 3;
+    this.state = {
```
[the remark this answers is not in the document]
mmmavis: `this.props.location` is an object and doesn't include a full URL as a property

e.g., 
image

If we really wanna use `new URL()` I guess we can do 

``` this.location = new URL(window.location)``` in `componentDidMount(){ ... }` ... thoughts?

File: components/join-us-modal/step.jsx (hunk 4 above)
Comment (mmmavis) [reply components/join-us-modal/step.jsx]: yup forgot to remove those 😂

File: components/navbar/navbar.scss (hunk 5 above)
Comment (mmmavis) [reply components/navbar/navbar.scss]: i looked it up online before i named this and it seemed like both were acceptable??? Anyways I'll rename it to indices.

File: components/join-us-modal/join-us-modal.jsx (hunk 3 above)
Comment (mmmavis) [reply components/join-us-modal/join-us-modal.jsx]: @gideonthomas ^ ?

Comment (gideonthomas) [reply components/join-us-modal/join-us-modal.jsx]: we'd have to change the api to support that feature. I think it's a lot of work for little value so I think that text currently suffices.

File: components/join-us-modal/form/create-fields.js (hunk 1 above)
Comment (mmmavis) [reply components/join-us-modal/form/create-fields.js]: @Pomax what data do I POST to this route? Is it just `email` along with `csrf_token` and `nonce`?

https://github.com/mozilla/network-pulse-api/blob/065dd3196b9598eb69a1d380e29b7eaf82d92b84/pulseapi/utility/newsletters/signup.py#L56-L83

File: components/join-us-modal/join-us-modal.jsx (hunk 7 above)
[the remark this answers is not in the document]
Comment (mmmavis) [reply components/join-us-modal/join-us-modal.jsx]: @Pomax ^

File: components/join-us-modal/form/create-fields.js (hunk 1 above)
Comment (Pomax) [reply components/join-us-modal/form/create-fields.js]: you shouldn't need the `nonce` for this route, and the payload should have an `email` field as well as a `source` field. The `source` field is used to say which URL the signup request came from so usually that's  `window.location.toString()`

File: components/join-us-modal/join-us-modal.jsx (hunk 7 above)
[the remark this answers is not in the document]
Comment (Pomax) [reply components/join-us-modal/join-us-modal.jsx]: Hm, where is the value for `this.props.location` generated? It kind of looks like it's already a `URL` object...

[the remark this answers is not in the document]
Comment (mmmavis) [reply components/join-us-modal/join-us-modal.jsx]: oh it's just a "built-in" prop that get passed by `react-router`. I guess it's more server side friendly?

File: package.json (hunk 6 above)
Comment (mmmavis) [reply package.json]: ```'components/**/*.js?(x)'``` worked!

**Comment by mmmavis** [comment]
@Pomax 
> Hm, there are no "back" buttons during the initial flow, which feels like they're missing.

We are okay with not having "back" button for now since everything can be edited later in `/myprofile`. @xmatthewx @natalieworth let me know if I misunderstood this.

**Comment by mmmavis** [comment]
@natalieworth this is ready for design review. But since this can only be tested locally either we set up a video call to go through the steps or we can pick a day to do this in the office together 😂

**Comment by natalieworth** [comment]
@mmmavis I should be better on tuesday and can go into office?:)

**Comment by natalieworth** [comment]
@mmmavis This is looking awesome. It's cool to see it live!!! I just screenshotted a couple of design things and included 1 question for the mobile at the end:

**1st screen:** change the "github/ google type to sentence case
Screen Shot 2019-07-03 at 9 24 50 AM

___________

**2nd screen:** In firefox, the -65/ error jumps to the left. I think hen you showed me it in Chrome, it was looking okay/ in the right spot. For the error screen, [could we change to this styling (with our brand red).](https://redpen.io/oy4a25cddc1141836a)
Screen Shot 2019-07-03 at 9 14 21 AM

___________

**3rd screen:** I think just the purple button styling on this has to be added and remove bold type on the button? Other than the button, everything else looks good on this page.
https://redpen.io/jpb788b0f44fcc1ef4

___________

**4th screen:** This looks good!

___________

Overall: Is it possible to center/ move the modal down so it's less close to the top of the page? Ideally if we could center/ move this down so there is more space at the top:
Screen Shot 2019-07-03 at 9 37 28 AM

The 'continue' button on click also seems to be changing to a different color:
button

Add spacing to the 'I'll do it later's'. Could we make the spacing equal between the continue button, the I'll do this later link, and the bottom of the modal?
Screen Shot 2019-07-03 at 9 43 11 AM



___________

**Mobile:** when I view as mobile in the firefox 'inspect element' space, it appears like the image below. Will it show up like this https://redpen.io/yj967ff9ec7cf24068 on actual mobile?
Screen Shot 2019-07-03 at 9 15 19 AM

**Comment by mmmavis** [comment]
@natalieworth 

> **1st screen:** change the "github/ google type to sentence case
Screen Shot 2019-07-03 at 9 24 50 AM

I'm trying to use existing brand style for that sentence. Out of all the type classes there are only two that are in gray. If `h6-heading` doesn't work are we okay with changing it to `body-small` then?

**Comment by mmmavis** [comment]
@natalieworth 

> 2nd screen: In firefox, the -65/ error jumps to the left. I think hen you showed me it in Chrome, it was looking okay/ in the right spot. For the error screen, could we change to this styling (with our brand red).

I can't reproduce this. Tried it on FF with different screen sizes. 😬 

For error color, let's do it as follow-up since this change will affect all form fields on the site.

**Comment by mmmavis** [comment]
> **Overall:** Is it possible to center/ move the modal down so it's less close to the top of the page? Ideally if we could center/ move this down so there is more space at the top:

Unfortunately it gets quite complicated code wise if we want to make modal always be vertically centered in any viewport. (It will probably have to involve with JS which usually is something we want to avoid as a style fix solution). Can we see this as a nice to have? I can file a ticket for this for us to revisit.
___________


> **Mobile:** when I view as mobile in the firefox 'inspect element' space, it appears like the image below. Will it show up like this https://redpen.io/yj967ff9ec7cf24068 on actual mobile?
Screen Shot 2019-07-03 at 9 15 19 AM

This shouldn't happen since mobile and desktop share exactly the same code. I wonder if this was related to the Cloudflare issue the Internet was having yesterday. @natalieworth can you test it again and if the bug is still there, let me know the exact steps you did so I can try reproduce it.

**Comment by mmmavis** [comment]
@natalieworth this is ready for design review again. I left some comments (see above) for things that I couldn't reproduce or stuff I'm suggesting to tackle as follow-ups. Let me know what you think!

**Comment by natalieworth** [comment]
@mmmavis For the first screen we can use the body small in sentence case? 

I’m also good with all your suggestions and P2’s/ follow up tickets. I can make a general design ticket to update forms across both sites. But all good from me! Ya I can’t reproduce the issues, so it may have been from the thing yesterday when it wasn’t working for us 😂

Let me know though if anything’s confusing and I can come in tomorrow or hop on zoom :)

File: js/service.js [hunk 8]
```diff
@@ -314,12 +316,22 @@ let Service = {
       return getDataFromURL(`${PULSE_API}/v2/myprofile/`);
     },
     put: function(updatedProfile) {
+      if (updatedProfile.newsletter) {
```
Pomax: we may want to do this only if we know the value changed from "not checked" to "checked". The way it is shouldn't cause any issues, but it might post signups that won't do anything when someone updates their profile...

components/join-us-modal/join-us-modal.jsx
```diff
@@ -0,0 +1,326 @@
...
+        isOpen={this.state.showModal}
+        onRequestClose={() => this.closeModal()}
+        shouldCloseOnOverlayClick={false}
+        className="join-us-modal"
+        overlayClassName="join-us-modal-overlay"
+      >
+        <button className="btn btn-close" onClick={() => this.closeModal()}>
+          <span className="sr-only">Close modal</span>
+        </button>
+        {!this.state.showConfirmation && this.renderProgressDots()}
```
Pomax: This is two separate template instructions both conditional on `showConfirmation`, let's refactor this to a single templating call that generates the right content based on whether or not that state variable is set.

components/join-us-modal/join-us-modal.jsx
```diff
@@ -0,0 +1,326 @@
...
+    );
+  }
+
+  renderConfirmation() {
+    return (
+      <div className="text-center">
+        <h2 className="h3-heading">Discover and Connect!</h2>
+        <p className="body-large">Thanks for signing up!</p>
+        <p>
+          Check out projects, connect with other members. Upload your projects
```
Pomax: was this supposed to be `[...] members, upload [...]`? As two separate sentences this text is a bit odd (both sentences right now list two things, but the first uses a comma, while the second uses an "or". That said, since it's really just a list of four things, we probably want one sentence that lists four things?)

components/join-us-modal/join-us-modal.jsx
```diff
@@ -0,0 +1,326 @@
...
+      showModal: false,
+      submitting: false,
+      showConfirmation: false
+    };
+  }
+
+  openModal() {
+    this.setState({ showModal: true }, () => {
+      // a hack to disable 'email' field
+      // because react-formbuilder doesn't allow setting a field as disabled
```
Pomax: hm, what are we using `disabled` for, in this case? (we can add prop passthrough to the formbuilder, but if we can achieve the same thing with a class, that's probably better?)

package.json
```diff
@@ -26,7 +26,7 @@
     "optimize:svg": "find assets -type f -name '*.svg' -print0 | xargs -0 -n 1 -P 6 svgo --multipass --pretty --enable=removeTitle",
     "optimize:png": "find assets -type f -name '*.png' -print0 | xargs -0 -n 1 -P 6 optipng",
     "watch:css": "chokidar 'components/**/*.scss' 'pages/**/*.scss' 'scss/**/*.scss' -c 'npm run build-client:scss'",
-    "watch:js": "chokidar 'app.jsx' 'main.jsx' 'js/**/*.js' 'components/**/*.jsx' 'pages/**/*.jsx' 'pages/**/*.js' 'webpack.config.js' -c 'npm run build-client:js'",
+    "watch:js": "chokidar 'app.jsx' 'main.jsx' 'js/**/*.js' 'components/**/*.js?(x)' 'pages/**/*.js?(x)' 'webpack.config.js' -c 'npm run build-client:js'",
```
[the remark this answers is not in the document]
mmmavis: I think glob doesn't use regular regex syntax. I followed their doc here: https://facelessuser.github.io/wcmatch/glob/

e.g.,
`'components/**/*.jsx?'`: only matches files that have four-character file extension prefixed with`jsx`

components/join-us-modal/join-us-modal.jsx
mmmavis: Thanks for catching this. I straight up copied and pasted the provided copy without thinking. Now I read it over I agree these sentences sound odd. @xmatthewx @natalieworth can you chime in here? We are talking about copy on this screen 

image

js/service.js
mmmavis: Gotcha but is there a way for us to verify user's newsletter subscription state? 🤔

components/join-us-modal/join-us-modal.jsx
gideonthomas: per @natalieworth change it to:
Check out projects or upload your own, connect with other members, or add more to your profile.

components/join-us-modal/join-us-modal.jsx
```diff
@@ -0,0 +1,336 @@
...
+    );
+  }
+
+  renderModalContent() {
+    if (this.state.showConfirmation) {
+      return this.renderConfirmation();
+    }
+
+    return (
+      <div>
```
mmmavis: the version of React Pulse uses doesn't have React fragment available so I had to wrap these elements in `<div>` instead

components/join-us-modal/join-us-modal.jsx
mmmavis: We want to display user's email in a non-editable field. ^ was the easiest solution I could come up with without touching formbuilder.

> we can add prop passthrough to the formbuilder, but if we can achieve the same thing with a class, that's probably better?

Can you elaborate on what you meant by using a class?

mmmavis:
Updated some things. Left comments on things I have questions about.

components/join-us-modal/join-us-modal.jsx
Pomax: a CSS class, but I guess that wouldn't prevent the field from being editable so... no dice on that!

js/service.js
Pomax: yeah that's a good point, I don't think there is right now =(

package.json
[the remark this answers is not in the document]
Pomax: ugh, keep forgetting that difference. Good point!

package.json
```diff
@@ -26,7 +26,7 @@
     "optimize:svg": "find assets -type f -name '*.svg' -print0 | xargs -0 -n 1 -P 6 svgo --multipass --pretty --enable=removeTitle",
```
Pomax: Windows does not allow single quotes in runtime arguments, so in order to work cross-platform these have to be "
```suggestion
    "optimize:svg": "find assets -type f -name \"*.svg\" -print0 | xargs -0 -n 1 -P 6 svgo --multipass --pretty --enable=removeTitle",
```

package.json
```diff
@@ -26,7 +26,7 @@
...
     "optimize:png": "find assets -type f -name '*.png' -print0 | xargs -0 -n 1 -P 6 optipng",
```
Pomax:
```suggestion
    "optimize:png": "find assets -type f -name \"*.png\" -print0 | xargs -0 -n 1 -P 6 optipng",
```

package.json
Pomax:
```suggestion
    "watch:js": "chokidar \"app.jsx\" \"main.jsx\" \"js/**/*.js\" \"components/**/*.js?(x)\" \"pages/**/*.js?(x)\" \"webpack.config.js\" -c \"npm run build-client:js\"",
```

package.json
```diff
@@ -26,7 +26,7 @@
...
     "watch:css": "chokidar 'components/**/*.scss' 'pages/**/*.scss' 'scss/**/*.scss' -c 'npm run build-client:scss'",
```
Pomax:
```suggestion
    "watch:css": "chokidar \"components/**/*.scss\" \"pages/**/*.scss\" \"scss/**/*.scss\" -c \"npm run build-client:scss\"",
```

PR #1101, opened 2019-06-12
New Search Homepage

youriwims:
# 👩‍🎨  READY FOR DESIGN REVIEW
### 👶 Child PRs 👉 #1102, #1103, #1104

Closes #1039 
Closes #1041 
Closes #1077

Related Tickets: #1039, #1041, #1059

_This PR takes care of the nav + search/dropdown style revisions & homepage redirect._

View Updates Here 👉 https://network-pulse-staging-pr-1101.herokuapp.com/

cc/ @xmatthewx @natalieworth @alanmoo @mmmavis

PR #1102, opened 2019-06-14
Search Bar Updates

youriwims:
## 🛑 THIS PR SHOULD 🚀 BEFORE #1103 & #1104 

**Related Prs**: #1103, #1104 
**Master PR**: #1101 

_This PR updates the search and dropdown styles according to design._

View Changes Here ➡️ https://network-pulse-staging-pr-1102.herokuapp.com/search

### DIFF
--- a/components/help-dropdown/help-dropdown.jsx
+++ b/components/help-dropdown/help-dropdown.jsx
@@ -23,7 +23,7 @@ export default class HelpDropdown extends React.Component {
     });
 
     return (
-      <select className={`custom-select w-100 ${this.props.specialClass}`} value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
+      <select className="custom-select w-100 help-dropdown" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
         <option value="">Help</option>
         {renderedHelpTypes}
       </select>
--- a/components/help-dropdown/help-dropdown.jsx
+++ b/components/help-dropdown/help-dropdown.jsx
@@ -23,7 +23,7 @@ export default class HelpDropdown extends React.Component {
     });
 
     return (
-      <select className="custom-select w-100" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
+      <select id={this.props.id} className="custom-select w-100" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
         <option value="">Help</option>
         {renderedHelpTypes}
       </select>
--- a/components/help-dropdown/help-dropdown.jsx
+++ b/components/help-dropdown/help-dropdown.jsx
@@ -23,7 +23,7 @@ export default class HelpDropdown extends React.Component {
     });
 
     return (
-      <select id={this.props.id} className="custom-select w-100" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
+      <select className={`custom-select w-100 ${this.props.specialClass}`} value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
         <option value="">Help</option>
         {renderedHelpTypes}
       </select>
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -126,7 +126,7 @@ class Search extends React.Component {
           placeholder="Search keywords, people, tags..."
           className="form-control"
         />
-        <span className="search-icon search small"></span>
+        <span className="search-icon search-glyph"></span>
       </div>
     </div>;
   }
@@ -152,6 +152,7 @@ class Search extends React.Component {
         <HelpDropdown 
           value={  this.state.helpType } 
           helpType={ (event) => this.handleHelpChange(event)} 
+          id="help-dropdown"
         />
       </div>
     </div>;
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -24,7 +24,7 @@ class Search extends React.Component {
     // Ticket filed on the 'react-debounce-input' repo https://github.com/nkbt/react-debounce-input/issues/65
     // In the meanwhile, we have to rely on document.querySelector(`#search-box`) to trigger input's focus() function
     document.querySelector(`#search-box`).focus();
-    document.querySelector(`.search-icon`).classList.add(`search-focus`);
+    document.querySelector(`.search-glyph-container`).classList.add(`search-focus`);
     this.inputFocusEvent();
   }
 
@@ -102,7 +102,7 @@ class Search extends React.Component {
 
   inputFocusEvent() {
     let input = document.querySelector(`#search-box`);
-    let icon = document.querySelector(`.search-icon`);
+    let icon = document.querySelector(`.search-glyph-container`);
 
     document.addEventListener("click", (event) => {
         if (event.target === input) {
@@ -126,7 +126,7 @@ class Search extends React.Component {
           placeholder="Search keywords, people, tags..."
           className="form-control"
         />
-        <span className="search-icon search-glyph"></span>
+        <span className="search-glyph-container search-glyph"></span>
       </div>
     </div>;
   }
@@ -152,7 +152,7 @@ class Search extends React.Component {
         <HelpDropdown 
           value={  this.state.helpType } 
           helpType={ (event) => this.handleHelpChange(event)} 
-          id="help-dropdown"
+          specialClass="help-dropdown"
         />
       </div>
     </div>;
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -115,7 +115,7 @@ class Search extends React.Component {
   }
 
   renderSearchBar() {
-    return <div className="d-flex align-items-center">
+    return <div className="d-flex align-items-center pr-lg-5 mr-lg-5">
       <div className="activated search-bar w-100">
         <DebounceInput id="search-box"
           value={this.state.keywordSearched}
--- a/pages/search/search.jsx
+++ b/pages/search/search.jsx
@@ -24,7 +24,7 @@ class Search extends React.Component {
     // Ticket filed on the 'react-debounce-input' repo https://github.com/nkbt/react-debounce-input/issues/65
     // In the meanwhile, we have to rely on document.querySelector(`#search-box`) to trigger input's focus() function
     document.querySelector(`#search-box`).focus();
-    document.querySelector(`.search-submit`).classList.add(`search-focus`);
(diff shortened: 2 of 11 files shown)


natalieworth:
@youriwims Functionality of this is great. I think only thing is the search icon gets cut off a little bit:
Screen Shot 2019-06-14 at 11 04 26 AM

natalieworth:
Styling-wise, I know Alan mentioned holding off on updating form styles. But wondering if we could update these styles when we start on the form tickets? I don't know if we should keep it blocked because the foundation site is different. Time-wise it's okay if we don't have this styled before All-hands though

youriwims:
I think we are updating the styles of the search + dropdown when we update the forms. @alanmoo can you confirm if that was the plan?

youriwims:
btw, just fixed the icon @natalieworth

alanmoo:
Yeah. They're all shared code, so it doesn't make sense to do one without the other. If this is the higher priority item and looks presentable without the updates, I'd encourage us to go with it.

natalieworth:
Thanks @alanmoo @youriwims So we can leave for now but update when we are able to get to the form updates

youriwims:
@natalieworth is this pr good to go?

natalieworth:
@youriwims That's great. I think last thing (and let me know if I should open a new ticket), is if it's possible to style the People/Projects links like this https://redpen.io/tra75dce779ed1fd27 (Could use the same styling as what is on the profile page with "about/favs": https://network-pulse-staging-pr-1102.herokuapp.com/profile/557

Screen Shot 2019-06-14 at 12 53 51 PM

alanmoo:
Great catch @natalieworth, those should indeed share styles!

youriwims:
Updated the tabs to use the same styling as what's on the profile page.

natalieworth:
@youriwims YES! Functionality looks/ feels great and showed to Matthew too. I think only thing left is just to update the styling. I know we originally were going to leave for now (but homepage is pretty important/ where everyone will be landing). :D

components/help-dropdown/help-dropdown.jsx
```diff
@@ -23,7 +23,7 @@ export default class HelpDropdown extends React.Component {
     });
 
     return (
-      <select className="custom-select w-100" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
+      <select id={this.props.id} className="custom-select w-100" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
```
youriwims: In case we reuse this component, let me know if it'd be better to add a special class instead of `id`.

scss/foundation-styles/glyphs.scss
```diff
@@ -58,9 +58,8 @@
   @include hoverGlyph("github-hover");
 }
 
-.search {
```
youriwims: There's already a `.search` class in the codebase  😬

scss/foundation-styles/mixins.scss
```diff
@@ -88,3 +88,30 @@
     }
   }
 }
+
+@mixin homepage-search-dropdown( $border-color, $dropdown ) {
```
youriwims: I created this mixin to differentiate between statement form elements on the homepage and regular ones throughout the site.

scss/mixin.scss
```diff
@@ -43,3 +43,30 @@
     }
   }
 }
+
+@mixin homepage-search-dropdown( $border-color, $home-dropdown ) {
```
youriwims: I created this mixin to differentiate between statement form elements on the homepage and regular ones throughout the site. Should I rename it?

scss/foundation-styles/mixins.scss
mmmavis: comment about the name of this mixin

- Let's remove `homepage` to make the name more generic (there aren't multiple versions of "search dropdown" style so we don't need to make the mixin name so specific)
- Can we rename it to`search-input-and-dropdown`? I know the name is loooong but `search-dropdown` makes me think that this mixin is for the help type dropdown only

scss/shared.scss
```diff
@@ -32,7 +32,6 @@
 }
 
 .custom-select {
-  border-radius: 0;
```
youriwims: This is covered by the [mixin](https://github.com/mozilla/network-pulse/pull/1102/files#diff-5a69ecdcb6159ed08581c1ee5d81d240R47).

scss/mixin.scss
```diff
@@ -43,3 +43,30 @@
...
+  @include scaleText(16px ,24px ,20px ,28px);
```
mmmavis: the comma-spacing format is weird here. it should be `, ` instead of ` ,` :)

scss/shared.scss
mmmavis: If we want to have one mixin for both search "input" and "dropdown" I would get rid of this `custom-dropdown` class and move the rules to the mixin to keep things all in one place and avoid rule duplicates.

scss/main.scss
```diff
@@ -164,13 +164,10 @@ fieldset {
...
 .form-control {
-  // TODO: Remove font-weight & add/@extend .body-large once Pulse forms are updated
-  font-weight: 300;
-  border-radius: 0;
+  @include homepage-search-dropdown( $gray-20, false );
 
-  &:focus,
-  &:active {
-    border-color: $light-blue;
+  &#search-box {
```
mmmavis: this ruleset doesn't have to nested in `input, textarea ... {}` as the selector is an `id`.

scss/mixin.scss
mmmavis: comment about the name of this mixin

- Let's remove `homepage` to make the name more generic (there aren't multiple versions of "search dropdown" style so we don't need to make the mixin name so specific)
- Can we rename it to `search-input-and-dropdown`? I know the name is loooong but `search-dropdown` makes me think that this mixin is for the help type dropdown only

pages/search/search.jsx
```diff
@@ -152,6 +152,7 @@ class Search extends React.Component {
         <HelpDropdown 
           value={  this.state.helpType } 
           helpType={ (event) => this.handleHelpChange(event)} 
+          id="help-dropdown"
```
mmmavis: Using `class` instead might be better

scss/main.scss
```diff
@@ -164,13 +164,10 @@ fieldset {
 input,
 textarea,
 .form-control {
-  // TODO: Remove font-weight & add/@extend .body-large once Pulse forms are updated
-  font-weight: 300;
-  border-radius: 0;
+  @include homepage-search-dropdown( $gray-20, false );
```
mmmavis: Using a "search input and dropdown" mixin on all `input`, `textarea`, and `.form-control` doesn't seem right to me. 

- if that mixin is meant for all `input`, `textarea`, and `.form-control` then the name should be more generic
- otherwise we shouldn't use that mixin here 

Since this PR is for getting the search input box look right, I would remove the `@include` line here to avoid making changes to other elements on the site. If we wanna update other form elements we should do that as a different PR.

scss/shared.scss
youriwims: ahh, good point 👍

pages/search/search.jsx
```diff
@@ -126,7 +126,7 @@ class Search extends React.Component {
           placeholder="Search keywords, people, tags..."
           className="form-control"
         />
-        <button className="btn search-submit search small"></button>
+        <span className="search-icon search-glyph"></span>
```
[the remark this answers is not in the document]
youriwims: yeah, def.

pages/search/search.jsx
youriwims: 🆗 🆒

scss/mixin.scss
youriwims: Lol I noticed that after the fact, thanks haha

scss/mixin.scss
youriwims: Cool cool, sounds good :)

scss/main.scss
youriwims: Yeah, the form elements will be updated in an upcoming ticket so I'll remove this. I just renamed the mixin to `search-input-and-dropdown`, but we can think of a new name between now and that ticket so we can reuse this mixin.

components/help-dropdown/help-dropdown.jsx
youriwims: https://github.com/mozilla/network-pulse/pull/1102/files#r297807293

pages/search/search.jsx
```diff
@@ -126,7 +126,7 @@ class Search extends React.Component {
           placeholder="Search keywords, people, tags..."
           className="form-control"
         />
-        <button className="btn search-submit search small"></button>
+        <span className="search-glyph-container search-glyph"></span>
```
mmmavis: I still find it odd to have these two classes on the same element but I guess that's just to align with how other existing glyphs are set up. Hopefully one day we can clean this up... (tbh i think the glyphs SCSS code from Foundation site is kinda outdated and needs to be updated)

pages/search/search.jsx
```diff
@@ -152,6 +152,7 @@ class Search extends React.Component {
         <HelpDropdown 
           value={  this.state.helpType } 
           helpType={ (event) => this.handleHelpChange(event)} 
+          specialClass="help-dropdown"
```
mmmavis: it's conventional to just call this `className`

mmmavis: We don't need this props. Here's the only place in the codebase that uses `<HelpDropdown />` so `help-dropdown` should just be the default class that comes with the HelpDropdown component.

components/help-dropdown/help-dropdown.jsx
```diff
@@ -23,7 +23,7 @@ export default class HelpDropdown extends React.Component {
     });
 
     return (
-      <select className="custom-select w-100" value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
+      <select className={`custom-select w-100 ${this.props.specialClass}`} value={ this.props.value } onChange={ (event) => {this.props.helpType(event);} }>
```
mmmavis: class `help-dropdown` should always be here. We shouldn't need `specialClass` prop to determine it.

pages/search/search.jsx
youriwims: Yeah, I agree this one is a bit weird. Let me know if you have any ideas on how to un-weird it 😅

youriwims: Feel free to open a ticket for the glyphs clean up, I wouldn't mind it tbh 😸

Issue #1039, opened 2019-04-26
Update Nav: Add button and other styles


natalieworth:
_Note: Nav update and Search component (#1041) are dependent and should ship together._

Remove the current navigation and update with the new button:

Signed out view
<img width="842" alt="Screen Shot 2019-04-26 at 12 30 07 AM" src="https://user-images.githubusercontent.com/14876567/56791019-e8debc00-67ba-11e9-940f-9cdd4eb9822b.png">

Signed in view (with account icon added if it's possible)
<img width="746" alt="Screen Shot 2019-04-26 at 12 30 29 AM" src="https://user-images.githubusercontent.com/14876567/56790999-e0868100-67ba-11e9-9afe-cc331983c2ba.png">

@youriwims adding the nav 'x', search, and profile icons here for when you update the navigation based on #1074 

[medium-icons.zip](https://github.com/mozilla/network-pulse/files/3234713/medium-icons.zip)
![58532058-4474de80-8199-11e9-9136-65cccbe7897b](https://user-images.githubusercontent.com/14876567/58596172-e8a56680-8227-11e9-9c04-359e98df74b5.png)

natalieworth:
[account-icon.svg.zip](https://github.com/mozilla/network-pulse/files/3120198/account-icon.svg.zip)

AdityaSrivast:
@natalieworth I would like to work on this issue. Can I take it up?

natalieworth:
Putting a hold on this issue as we still have to flush out some of the menu UX for mobile

natalieworth:
@xmatthewx Just noting here, that I think we still may need some navigation for mobile so users can access their account? I've mocked this up and included some suggestions in the 'signed in' menu view: https://redpen.io/p/gc34eb3beb2157b54e

<img width="664" alt="Screen Shot 2019-04-29 at 7 32 33 PM 1" src="https://user-images.githubusercontent.com/14876567/56938269-c64bec00-6ab5-11e9-8a65-b5a0a82610f0.png">

xmatthewx:
Yes. Pulse does currently have nav drawer on mobile. Your revised UI looks good. For items, let's just do Profile, Edit profile, Favs, Signout. (I like how you provided some shortcuts.)

natalieworth:
Mobile menu's have been updated for this/ unblocking.

Signed out view https://redpen.io/xsf5b7302d480d9b5d
Signed in view https://redpen.io/pb2fcfb9e1a16cc0c3

[...]

PR #1103, opened 2019-06-17
Nav updates

youriwims:
## 💚 READY TO MERGE 💚 

**Related Ticket**: #1039 
**Related Prs**: #1102, #1104
**Master PR**: #1101 

_This PR updates the nav styles. The nav is only visible on mobile and has two screens: 'SignIn/SignUp' request if you're ***not*** logged in and nav items ***if you are*** logged in. Please Note: You ***will not*** see the search page until the redirect PR (#1104)._

View Updates Here 👉 https://network-pulse-staging-pr-1103.herokuapp.com/featured


youriwims:
@natalieworth I updated the vertical spacing above the sign-in request content to be `120px` instead of the `160px` in the mock because it looked like too much spacing when I tested it out on multiple mobile devices (iPhone 5, iPhone 6/7/8, iPhone 6/7/8 Plus,  iPhone X/XS & Pixel  2). Let me know what you think.

Screen Shot 2019-06-25 at 11 56 23 AM

natalieworth:
Niiceee yes @youriwims

components/navbar/navbar.scss
```diff
@@ -1,142 +1,136 @@
...
+      }
     }
   }
 
-  .pinned {
-    position: absolute;
-    top: 0;
-    right: 0;
+  .signupin-request {
+    margin-top: 120px;
```
youriwims: Reference: https://github.com/mozilla/network-pulse/pull/1103#issuecomment-505510623

scss/foundation-styles/buttons.scss
```diff
@@ -83,39 +83,39 @@
...
-
-    @include hover-focus-active {
-      &::after {
-        filter: brightness(0);
-      }
-    }
-  }
-}
+// // append external link glyph
+// .btn-secondary[href*="//"]:not([href*="api.mozillapulse.org"]) {
```
youriwims: This attempted to append a glyph to the SignIn / SignOut button on mobile nav when you're logged out. What would be the correct url to add here to avoid this so I can uncomment these lines 😅 ?

mmmavis:
@youriwims what's the ticket # this PR is fixing?

youriwims:
#1039 @mmmavis

components/navbar/navbar.scss
```diff
@@ -1,142 +1,136 @@
-#burger {
```
mmmavis: If certain pixel values in this file need to match I will turn them into variables.

e.g., `28px`, `12.5px` (should be changed to an integer number)

scss/foundation-styles/buttons.scss
mmmavis: maybe we should toggle `signupin` class on the mobile SignIn / SignOut button based on user's login status?

components/navbar/navbar.jsx
```diff
@@ -44,13 +41,15 @@ class NavBar extends React.Component {
   renderName(classes) {
     // don't show anything until we verify this user's loggedin status
     if (user.loggedin === undefined) return null;
+    let classNames = "signupin d-none d-md-flex align-items-md-center";
```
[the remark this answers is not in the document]
youriwims: I'll rename this, but I actually ended up removing the `classnames` import in the next PR since it wasn't being used here.

components/navbar/navbar.jsx
```diff
@@ -105,12 +108,13 @@ class NavBar extends React.Component {
...
+  renderNavContent() {
     let classes = classNames(
-      `nav-link-list list-unstyled ml-lg-4 mt-md-3 mt-lg-0 mb-0`,
+      `nav-link-list px-3 hidden-lg-up dark-theme`,
       {
-        "d-flex": this.state.burgerActive,
-        "flex-column": this.state.burgerActive
+        "list-unstyled": user.loggedin,
+        "d-flex": user.loggedin,
+        "flex-column": user.loggedin
```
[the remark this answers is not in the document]
youriwims: I double checked and you're right 😅

components/navbar/navbar.scss
youriwims: okay, will do 👍

scss/foundation-styles/buttons.scss
mmmavis: So we can't remove this entire chunk of SCSS since we will need it for the "Get involved" button on the entry page (see mockup of the new design). Since Pulse is not powered by CMS like the Foundation site, I think we can do this in a more manual way to avoid crazy complicated SCSS selector

Change 
```.btn-secondary[href*="//"]:not([href*="api.mozillapulse.org"])```
to 
```.btn-secondary.btn-outbound-link```

And when we get to the entry page ticket we can "add" the glyph to the "Get invovled" button by assigning the right classes to it (e.g., `btn btn-secondary btn-outbound-link`)

Comment (youriwims) [reply scss/foundation-styles/buttons.scss]: Okay, perfect. Thanks!

**Comment by youriwims** [comment]
@mmmavis this is ready for re-review! :)

File: components/navbar/navbar.scss [hunk 6]
```diff
@@ -1,142 +1,132 @@
...
-  margin-left: 1rem;
-  border: 0;
-  background: url(/assets/svg/hamburger.svg) no-repeat;
-  width: 31px;
-  height: 39px;
-}
+#main-nav-wrapper {
+  .burger {
+    margin-right: $value-28px;
```
mmmavis: So since these variables are only used for things in `#main-nav-wrapper` we can declare the variables within the scope of its rule set.

```scss
#main-nav-wrapper {
  $burger-width: 28px;
  $burger-height: 28px;
  $burger-bar-offset: 13px;

  .burger {
    ...
    ...

}

```

PR #1104, opened 2019-06-26
Homepage redirect

youriwims:
## 💚 READY FOR REVIEW 💚

**Related Prs**: #1102, #1103 
**Master PR**: #1101 

_This PR redirects the homepage to the search component instead of Featured and adds the 'Learn More' tagline._

View Updates Here  👉  https://network-pulse-staging-pr-1104.herokuapp.com/

### DIFF
--- a/components/help-dropdown/help-dropdown.jsx
+++ b/components/help-dropdown/help-dropdown.jsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import classNames from 'classnames';
 import Service from '../../js/service';
 
 export default class HelpDropdown extends React.Component {
--- a/main.jsx
+++ b/main.jsx
@@ -23,17 +23,6 @@ import NotFound from "./pages/not-found.jsx";
 import Navbar from "./components/navbar/navbar.jsx";
 import Footer from "./components/footer/footer.jsx";
 
-const Featured = () => {
-  return (
-    <div>
-      <Helmet>
-        <title>Featured</title>
-      </Helmet>
-      <ProjectLoader featured={`True`} />
-    </div>
-  );
-};
-
 const Latest = () => {
   return (
     <div>
@@ -70,7 +59,6 @@ const Routes = () => (
     <Route exact path="/" component={Search} />
     <Route exact path="/search" component={Search} />
     <Route path="/search/:tab" component={Search} />
-    <Route path="/featured" component={Featured} />
     <Route path="/latest" component={Latest} />
     <Route path="/favs" component={Bookmarks} />
     <Route exact path="/issues" component={Issues} />
--- a/main.jsx
+++ b/main.jsx
@@ -24,36 +24,11 @@ import Navbar from "./components/navbar/navbar.jsx";
 import Footer from "./components/footer/footer.jsx";
 
 const Featured = () => {
-  let handleOnClick = function() {
-    ReactGA.event({
-      category: `Browse`,
-      action: `About learn more tap`,
-      label: `Tagline learn more link`
-    });
-  };
-
-  let learnMore = env.LEARN_MORE_LINK ? (
-    <span>
-      <a
-        href={env.LEARN_MORE_LINK}
-        id="learn-more"
-        onClick={() => handleOnClick()}
-      >
-        Learn more
-      </a>.
-    </span>
-  ) : null;
-
   return (
     <div>
       <Helmet>
         <title>Featured</title>
       </Helmet>
-      <div className="row">
-        <h2 className="mb-4 h2-heading col-12 col-md-10 col-lg-8">
-          Discover & collaborate on projects for a healthy internet. {learnMore}
-        </h2>
-      </div>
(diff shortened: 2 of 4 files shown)


components/help-dropdown/help-dropdown.jsx
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import classNames from 'classnames';
```
youriwims: This isn't related to this pr. It's just not being used so I removed it when I came across it.

main.jsx
```diff
@@ -24,36 +24,11 @@ import Navbar from "./components/navbar/navbar.jsx";
 import Footer from "./components/footer/footer.jsx";
 
 const Featured = () => {
```
youriwims: I kept the Featured component & `<Route>` in case we wanted to utilize it in the future. 🤷‍♀

pages/search/search.jsx
```diff
@@ -56,7 +57,7 @@ class Search extends React.Component {
   updateBrowserHistory() {
     let keywordSearched = this.state.keywordSearched;
     let helpType = this.state.helpType;
-    let location = { pathname: this.props.location.pathname };
+    let location = { pathname: "/search" };
```
youriwims: ...wanted to make sure old link work which currently follow the `/search?...` path. Let me know if there's a better way to do this 🤔

mmmavis:
Are we keeping nav items (Features, Latest, etc)?

image

main.jsx
mmmavis: We are unlikely going to _toggle_ the Featured component in the future so let's remove its related code if it's not in use.

mmmavis:
Question: do we wanna keep the `/search` route and let it function like the homepage (`/`) OR we want it so that when user's hit `/search` we redirect the URL to `/` (along with query params)?

youriwims:
> Are we keeping nav items (Features, Latest, etc)?
> 
> image

@mmmavis as far as I know, we're not keeping them. Maybe @natalieworth & @xmatthewx have more details 🤷‍♀

youriwims:
> Question: do we wanna keep the `/search` route and let it function like the homepage (`/`) OR we want it so that when user's hit `/search` we redirect the URL to `/` (along with query params)?

It was my understanding that `/search` _is_ the new homepage, but if I understand your question, I think the second. `/search` should redirect to `/` (which I think I need to do now that you bring it up 🙈) & `/` should show the `search` component. **_But_** for old links to work that have `/search`, when a user searches using dropdown or search bar we need to add `/search` back so the final path looks like ➡️ `/search?param-goes-here`.

youriwims:
@mmmavis ⬆️ 😬

mmmavis:
@youriwims 

> It was my understanding that `/search` _is_ the new homepage, but if I understand your question, I think the second. `/search` should redirect to `/` (which I think I need to do now that you bring it up 🙈) & `/` should show the `search` component. **_But_** for old links to work that have `/search`, when a user searches using dropdown or search bar we need to add `/search` back so the final path looks like ➡️ `/search?param-goes-here`. 

Cool cool. Can you update your PR so that the `/search` to `/` redirect happens in `main.jsx`? You will need to use `<Redirect>` from `react-router`. Link to doc [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md).

youriwims:
without a doubt 👍

xmatthewx:
Yes, we are dropping the Nav items replaced by a streamlined homepage.

youriwims:
> Cool cool. Can you update your PR so that the `/search` to `/` redirect happens in `main.jsx`? You will need to use `<Redirect>` from `react-router`. Link to doc [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md).

Sorry, I spoke too soon @mmmavis. We actually need to keep the `/search` `<Route>` since when a user uses search bar or dropdown the url updates to `/search?param-bla-bla` so we need that path not to redirect to `/` 😬

youriwims:
Fyi: Pulse api env variables needs to updated @gideonthomas

main.jsx
```diff
@@ -92,17 +56,18 @@ const Tag = router => {
 
 const Routes = () => (
   <Switch>
-    <Route exact path="/" render={() => <Redirect to="/featured" />} />
-    <Route path="/featured" component={Featured} />
+    <Route exact path="/" component={Search} />
+    <Route path="/search/:tab" render={(props) => <Redirect to={`/${props.match.params.tab}${props.location.search}`}/>}/>
```
alanmoo: This feels a bit opaque- why is it a Route with a Redirect in the action, instead of a Redirect component right away? Something to do with passing parameters?

youriwims: Yeah, it strips out the query params.

youriwims:
@alanmoo we didn't test that use case, but good point. Hopefully, what I just pushed up should fix that.

PR #1106, opened 2019-07-02
More sign-in specificity

alanmoo:

### DIFF
--- a/pages/add/add.jsx
+++ b/pages/add/add.jsx
@@ -276,7 +276,7 @@ class Add extends React.Component {
         href={user.getLoginURL(utility.getCurrentURL())}
         onClick={event => this.handleSignInBtnClick(event)}
       >
-        Sign in with Google
+        Sign in with Github or Google
       </a>
     );
     let additionalMessage;


youriwims:
The button looks like this because it's pointing to an external link and is attempting to append an 'external link' glyph/svg that doesn't yet exist. We have a short-term fix in the new homepage PRs that'll fix this once it lands on master. Long term fix will happen in the entry page ticket/PR.
cc: @xmatthewx @alanmoo @mmmavis

mmmavis:
image

There's supposed to be a outbound link looking glyph but I think there's a bug in the file path that's why the glyph is not showing

image

alanmoo:
Yup; I was going to file that as a separate issue if it's not known already but I wanted to get the words fix in.

mmmavis:
Yea let's get this merged in first!

xmatthewx:
@jenntbeard - the login button should soon make it clear you can use either google or github.

alanmoo:
(Probably tomorrow- the internet is having some struggles today)

PR #1111, opened 2019-07-12
Add sign up form to footer

### LINKED ISSUE #1063: Add sign-up section to footer
Continuing the work from #1054 — adding the sign-up component to the footer to match the foundation.mozilla.org site.

cc @youriwims @mmmavis

mmmavis:
Closes #1063


components/newsletter-sign-up/newsletter-sign-up.jsx
```diff
@@ -0,0 +1,288 @@
+import React from "react";
```
mmmavis: Pretty much the same as https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/components/join/join.jsx with some minor changes (doesn't render name fields as Pulse doesn't need it)

components/newsletter-sign-up/newsletter-sign-up.jsx
```diff
@@ -0,0 +1,288 @@
...
+    );
+  }
+
+  /**
+   * Submit the user's data to the API server.
+   */
+  submitDataToApi() {
+    this.setState({ apiSubmitted: true });
+
+    return Service.signUp(this.email.value, window.location.toString())
```
mmmavis: Utilizing Pulse's existing `service.js` instead of https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/components/join/join.jsx#L51-L93

components/newsletter-sign-up/newsletter-sign-up.jsx
```diff
@@ -0,0 +1,288 @@
...
+  apiIsDone() {
+    return (
+      this.state.apiSubmitted && (this.state.apiSuccess || this.state.apiFailed)
+    );
+  }
+
+  /**
+   * Submit the user's data to the API server.
+   */
+  submitDataToApi() {
```
mmmavis: Utilizing Pulse's existing `service.js` instead of https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/components/join/join.jsx#L51-L93

js/service.js
```diff
@@ -178,6 +178,7 @@ function updateStoredData(requestType = `POST`, endpointRoute, data = {}) {
 
         request.setRequestHeader(`X-CSRFToken`, data.csrfmiddlewaretoken);
         request.setRequestHeader(`Content-Type`, `application/json`);
+        //   xhr.setRequestHeader(`X-Requested-With`, `XMLHttpRequest`);
```
mmmavis: To code reviewer:
I saw this on Foundation site but I'm not sure if we need this for Pulse? Thoughts?

mmmavis:
@gideonthomas assigning you as code reviewer since others are on PTO this week. If you don't have bandwidth I guess we can leave the review until next week.

components/footer/footer.jsx
```diff
@@ -76,26 +77,30 @@ export default class Footer extends React.Component {
...
-              <div className="my-5 my-md-0 d-md-flex align-items-md-end h-100">
-                <a href="https://foundation.mozilla.org" className="logo">
-                  <img
-                    src="../assets/svg/mozilla-block-white.svg"
-                    alt="Mozilla Foundation Home Page"
-                  />
+            <div className="col-md-5 col-lg-6">
+              <NewsletterSignUp
+                buttonPosition="side"
+                ctaHeader="Want to get smarter about your online life?"
```
gideonthomas: Per https://github.com/mozilla/foundation.mozilla.org/issues/3161 let's remove this and stick to the default

components/newsletter-sign-up/newsletter-sign-up.jsx
```diff
@@ -0,0 +1,289 @@
...
+        this.apiSubmissionFailure(reason);
+      });
+  }
+
+  /**
+   * Performs very simple validation for emails.
+   * @param {string} input the string that should be validated as an email address
+   * @returns {boolean} true if the input is a legal-enough email address, else false
+   */
+  validatesAsEmail(input) {
```
mmmavis: Should we replace this with dummy form field validation?

js/service.js
gideonthomas: not sure tbh, this header is generally used for CSRF protection with CORS. If this request is working without this header we can leave it as it is.

components/footer/footer.jsx
```diff
@@ -76,26 +77,30 @@ export default class Footer extends React.Component {
...
+                ctaDescription="Sign up for our Mozilla newsletter!"
```
gideonthomas: same thing here, let's remove and stick to the updated default

components/newsletter-sign-up/newsletter-sign-up.jsx
gideonthomas: honestly up to you! We can also file a follow-up and get Pomax's opinion coz I'm not 100% sure it's necessarily the right way to do it.

mmmavis:
@gideonthomas I will test it again but last time I checked this only worked locally with local pulse api (staging/prod are all blocking localhost and review apps)

mmmavis:
> @mmmavis Design looks great. Just wondering though if it's okay if the content is different for each on the email sign up area. I noticed the Foundation site's content was updated?

@natalieworth I copied the text in your Sketch file. If that's not the actual content we want to have, do you know who we should talk to to get the right content?

gideonthomas:
@mmmavis so, I wasn't logged into pulse when I tried. So might wanna try it that way

gideonthomas:
@mmmavis re: the content question you have, I addressed that in my PR review (i.e. use the default content we have in the NewsletterSignup component and don't override the ctaHeader and ctaDescription props). That way, we will have the same content as on the foundation site.

mmmavis:
@gideonthomas 

> @mmmavis so, I wasn't logged into pulse when I tried. So might wanna try it that way

Thanks for catching the 403 error for unauthenticated users.

https://github.com/mozilla/network-pulse-api#get-apipulsenonce

All `POST` requests must be accompanied with the CSRF token but we can only get CSRF token for authenticated users (via the `/nonce` route). I guess we need to add a route to API for returning CSRF token for everyone including people who aren't logged in?

gideonthomas:
Hmm, so that sucks. I don't think we can make the CSRF/nonce route accessible without authentication because that will defeat the purpose of CSRF.

The only thing I can think of is to create a route on the pulse front-end server to fetch a CSRF token from the API server (we'll need to make the API server "trust" the front-end server for that route) and pass that on to the front-end to use. This is a sizable amount of work.

@patjouk any chance you know how to fix this? The issue is that we have a newsletter signup form on the pulse front-end that needs to post to the pulse api. Currently for any POSTs on the API you need a CSRF token and nonce value. To obtain these values, you have to be logged into the API.

@xmatthewx we have a bit of an issue here -- currently it's not straightforward to allow newsletter signups for people who are not logged into pulse. Is it ok to compromise on that for now (remove the signup piece in the footer if you aren't signed in)? Or is that a must-have?

xmatthewx:
We can suppress it for people who are not signed in.

mmmavis:
Will file a follow-up ticket regarding newsletter sign up CSRF token discussion.

PR #1112, opened 2019-07-15
Entry page

youriwims:


youriwims:
fyi: I had to remove the `.container` class in `main.jsx` in order to have the full width _Help needed_ section 😕. I ended up adding that class back on each individual page's parent container to allow more flexibility in page design/layout and fix any 🐛 before they 👻.Here's the specific commit for that 👉 https://github.com/mozilla/network-pulse/pull/1112/commits/1afe6758d3fe9cc70e14d013a8d545a4a34e3425

Also, @natalieworth are the social media buttons supposed to be sticky?

/cc @mmmavis

natalieworth:
@youriwims Ya, sticky if possible :)

youriwims:
See #1130

Issue #1116, opened 2019-07-16
Moderation needs to be updated


youriwims:
**Describe the bug**
Moderation was left behind in all the UI refresh fun 🤡 

**To Reproduce**
Visit https://www.mozillapulse.org/moderation

**Expected behavior**
Components should follow updated styling as documented in the [UI Refresh](https://github.com/mozilla/network-pulse/projects/10)

**Screenshots**
![Screenshot_2019-07-16 Mozilla Network Pulse](https://user-images.githubusercontent.com/36484608/61336920-3fc7c080-a801-11e9-9319-8c010f3aaa30.png)

//cc @mmmavis  @alanmoo @natalieworth @xmatthewx

xmatthewx:
Good catch. Moderation can be ugly, as long as it's functional.

PR #1124, opened 2019-07-31
Moderation updates

youriwims:
Closes: #1116 

View Changes 👉  https://network-pulse-staging-pr-1124.herokuapp.com/moderation

TODO: _Create follow up ticket for dropdown style alignment_

### DIFF
--- a/components/project-card/meta/issues-and-tags.jsx
+++ b/components/project-card/meta/issues-and-tags.jsx
@@ -5,29 +5,52 @@ import classNames from "classnames";
 import Utility from "../../../js/utility.js";
 
 const IssuesAndTags = props => {
-  let issues = props.issues.map(issue => {
-    return (
+  let issues = [];
+  let tags = [];
+
+  props.issues.map(issue => {
+
+    // Creates individual issue
+    issue = (
       <Link
-        to={`/issues/${Utility.getUriPathFromIssueName(issue)}`}
-        className="btn btn-xs btn-tag"
-        key={issue}
-      >
-        {issue}
-      </Link>
+      to={`/issues/${Utility.getUriPathFromIssueName(issue)}`}
+      key={issue}
+    >
+      {issue}
+    </Link>
     );
+
+    // Pushes individual issue followed by comma
+    issues.push(issue);
+    issues.push(`, `);
   });
 
-  let tags = props.tags.map(tag => {
-    return (
+  // If issues array full length, remove last comma
+  if (issues.length) {
+    issues.pop();
+  }
+
+  props.tags.map(tag => {
+
+    // Creates individual tag
+    tag = (
       <Link
         to={`/tags/${encodeURIComponent(tag)}`}
-        className="btn btn-xs btn-tag"
         key={tag}
       >
-        {tag}
+        #{tag}
       </Link>
     );
+    
+    // Adds comma before tag since issues.arr will never end in a comma
+    tags.push(`, `);
+    tags.push(tag);
   });
+
+  // However, if issues is empty, remove the first comma in tags.arr
+  if (issues.length == 0) {
+    tags.shift();
+  }
 
   let classnames = classNames(`issues-and-tags`, props.className);
 
--- a/components/project-card/meta/issues-and-tags.jsx
+++ b/components/project-card/meta/issues-and-tags.jsx
@@ -5,59 +5,38 @@ import classNames from "classnames";
 import Utility from "../../../js/utility.js";
 
 const IssuesAndTags = props => {
-  let issues = [];
-  let tags = [];
+  if (props.issues.length + props.tags.length === 0) return null;
 
-  props.issues.map(issue => {
-
-    // Creates individual issue
-    issue = (
+  let issues = props.issues.map(issue => {
+    return (
       <Link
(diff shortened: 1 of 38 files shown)


scss/overrides/_buttons.scss
```diff
@@ -28,3 +28,23 @@
     }
   }
 }
+
+a.btn {
```
mmmavis: Wrong file. These aren't overrides for Foundation site style guide elements.

mmmavis: Is `.btn-tag` a modified version on top of `.btn-secondary`?

pages/moderation.jsx
```diff
@@ -28,7 +28,7 @@ class Moderation extends React.Component {
   }
 
   render() {
-    if (!user.moderator) return <NotFound />;
+    //if (!user.moderator) return <NotFound />;
```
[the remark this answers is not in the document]
youriwims: Is there a way to view moderation without commenting that out?...that was the only way I could view it 😅

scss/overrides/_buttons.scss
youriwims: oops, i'll move that to a different file then. nope, `.btn-tag` is the new capsule-like button that Nat is trying to add to the style guide.

pages/moderation.jsx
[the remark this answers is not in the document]
mmmavis: Yea if your account has moderator rights... let me check with Giddy and ask you to the group. On localhost you should be able to do that yourself since you are the superuser.

components/project-card/meta/issues-and-tags.jsx
```diff
@@ -5,30 +5,53 @@ import classNames from "classnames";
 import Utility from "../../../js/utility.js";
 
 const IssuesAndTags = props => {
```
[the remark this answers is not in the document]
youriwims: > Few notes:
> 
>     1. `array.map()` returns a new array already so you shouldn't need to do `.push()` manually
>     2. we can return early if both the issues and tags lists are empty

Ahhh, good point.

components/project-card/meta/issues-and-tags.jsx
```diff
@@ -20,21 +21,22 @@ const IssuesAndTags = props => {
...
-        {tag}
+        #{tag}
       </Link>
     );
   });
 
   let classnames = classNames(`issues-and-tags`, props.className);
+  let issues_and_tags = issues.concat(tags);
+
+  issues_and_tags = issues_and_tags.reduce((accu, curr) => [accu, ", ", curr]);
```
mmmavis: @youriwims do you wanna change this to something easier to understand?

youriwims: Nope—I'm good with this now 😸

pages/moderation-search/moderation-search.scss
```diff
@@ -8,6 +8,28 @@
   height: 100%;
 
   .Select-control {
-    border-color: $input-border-color;
+    border-color: $black;
```
mmmavis: Are these using the same rules as the ones we apply to homepage's select dropdown? If so we should think of a way for them to share the same rule set instead of having duplicate.

scss/main.scss
```diff
@@ -212,6 +201,26 @@ fieldset {
   }
 }
 
+a.btn {
```
mmmavis: Does the selector need to be this specific (`a.btn.btn-tag`) for the right rules to kick in? I assume some rules should have inherited from `.btn` already?

pages/moderation-search/moderation-search.scss
youriwims: Yeah, it looks like the dropdown is mainly sharing `height`,`border-radius`, and `border-color`. You think I should create a mixin for them?

scss/main.scss
youriwims: Yeah, it's mainly to _not_ inherit the `700` `font-weight` that's on `a.btn` from foundation button styles.

mmmavis: Ah if that's the case can we just leave `font-weight: 400;` in `a.btn.btn-tag` and extract everything else to be in just `.btn-tag { }`? My concern is if everything is in `a.btn.btn-tag { }` now if we ever have to override any rules there in the future it'll get extra hard as the selector is very specific already.

Hopefully once major updates are done for Pulse we will have a chance to revisit all new styles introduced in Pulse and tidy up our SCSS to get rid of as many as overrides as we could.

pages/moderation-search/moderation-search.scss
mmmavis: Ah gotcha. For this to not become a blocker - as a followup we should align styling of all dropdowns on Pulse.

youriwims: Ya, good idea.

scss/main.scss
youriwims: Trueee, yeah I can def do that. 👍

youriwims: > can we just leave font-weight: 400; in a.btn.btn-tag

btw, I put ☝️  the button override styles since it's overriding foundation styles.

PR #1125, opened 2019-07-31
Main container fix

### LINKED ISSUE #1122: Main content needs more space
**Describe the bug**
The footer is overlapping main content [when a user is signed in and out] at mobile. Then, at desktop there's a lack of space between main content and footer.

**To Reproduce**
1. Visit any page on [staging](https://pulse.mofostaging.net/)
2. Scroll down to the footer
4. See either overlapping content _OR_ lack of space between main content + footer

**Expected behavior**
No footer overlap and at least 48px between footer + main content.

**Screenshots**
Screen Shot 2019-07-30 at 6 59 34 PM
Screen Shot 2019-07-30 at 7 06 23 PM

cc: @mmmavis

youriwims:
Closes #1122

### DIFF
--- a/components/footer/footer.scss
+++ b/components/footer/footer.scss
@@ -1,10 +1,3 @@
-footer {
-  position: absolute;
-  right: 0;
-  bottom: 0;
-  left: 0;
-}
-
 .site-footer {
   background-color: $black;
 
--- a/main.jsx
+++ b/main.jsx
@@ -118,7 +118,7 @@ class Main extends React.Component {
         />
         <JoinUsModalWithRouter />
         <NavbarWithRouter />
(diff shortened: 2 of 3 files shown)


youriwims:
@mmmavis I just made it `60vh`, let me know if that's better or if you have any other ideas because I don't really like the footer jumping tbh 😕 .

mmmavis:
@youriwims yea this is probably the reason why I went with the not-so-straightforward absolute positioning approach in the first place but that does require us to continuously update the SCSS (`padding-bottom` on `#main`) to make sure things look good on every page.

I'm fine with the min height approach here until we figure out a better way. Let's get this in to fix the overlapping footer on content bug on prod.

Issue #1117, opened 2019-07-17
Add the 'X' back to the search to clear results


natalieworth:
Just making a note to add the 'x' clear functionality back to mobile as a P2 when Youri returns.

cc @youriwims @mmmavis @alanmoo

mmmavis:
@natalieworth can you provide comp and the 'x' icon file? I can't recall how Pulse used to look like 😂

mmmavis:
For simplicity and consistency, wondering if we should do this for all screen sizes / devices other than just mobile?

/cc @alanmoo (context: the Search component doesn't know what screen size / device it's on. If we wanna target this just for mobile we'll have to write extra JS and logic to detect container size / device type to conditionally display the `x` control)

xmatthewx:
is it a bad practice to display or not with css based on screensize?

mmmavis:
I wouldn't say it's bad practice since it's usually case by case + how the component was set up initially. I will know more once I see the comps. Also by "mobile" do we mean when the page is displayed on viewport under 768px wide? Or on a mobile device?

alanmoo:
> is it a bad practice to display or not with css based on screensize?

It can be kind of arbitrary. You could have a narrow desktop window when trying to view a site side-by-side, or a big tablet.

natalieworth:
@mmmavis Sorry for delay. The previous version of pulse had the 'x' icon across all screen sizes:
<img width="1219" alt="Screen Shot 2019-07-25 at 2 45 55 PM" src="https://user-images.githubusercontent.com/14876567/61985510-64205b80-afbe-11e9-9e83-b2e92c772c32.png">

I'm okay with maintaining that functionality. 


----


### Assets
I have added the 'x' icon asset for this here:
[clear-x-icon.zip](https://github.com/mozilla/network-pulse/files/3437580/clear-x-icon.zip)

----

### Designs

Design for desktop is here:
https://redpen.io/efddc1be56ae5ffb13

Design for mobile is here:
https://redpen.io/ta03a3b0cb3d17d4bf


Let me know if you have any questions @mmmavis or I can also hop on a call to go over the functionality if it doesn't make sense. I'll also be in next week too depending on your schedule.

cc @youriwims @alanmoo

youriwims:
@natalieworth should the 'x' icon be black when when search isn't active?

natalieworth:
@youriwims ya, thanks

mmmavis:
[...]

PR #1126, opened 2019-08-01
Add x button to homepage search input box

mmmavis:
Related issue: #1117 

@natalieworth I made the `x` icon to be `16px * 16px` across all screen sizes to stick with other existing icon style on the site. On mockup it's 16x16 on desktop and 14x14 on mobile (also note that 14x14 might be too small for a control on mobile)

### DIFF
--- a/components/search-tab-group/search-tab-group.jsx
+++ b/components/search-tab-group/search-tab-group.jsx
@@ -9,26 +9,31 @@ import ProfileLoader from "../profile-loader/profile-loader.jsx";
 const TAB_NAMES = [`people`, `projects`];
 const DEFAULT_TAB_NAME = `projects`;
 
-class SearchTabGroup extends React.Component {
+// Extends React.PureComponent instead of React.Component as we want
+// this component to re-render only when there's change in its props or state.
+// Note that React.PureComponent runs *shallow comparison* to determine
+// if the componet should be re-rendered or not. So make sure
+// SearchTabGroup's props and state doesn't contain anything mutatable otherwise
+// we will run into issues.
+class SearchTabGroup extends React.PureComponent {
   constructor(props) {
     super(props);
 
+    this.availableTabs = TAB_NAMES;
     this.state = this.getInitialState(props);
   }
 
   getInitialState(props) {
-    let availableTabs = TAB_NAMES;
     let activeTab;
     let states = {
-      activeTab,
-      availableTabs
+      activeTab
     };
 
     // when there are no available tabs to show
-    if (availableTabs.length === 0) return states;
+    if (this.availableTabs.length === 0) return states;
 
     // the tab we are trying to load has content and can be rendered
-    if (availableTabs.indexOf(props.activeTab) >= 0) {
+    if (this.availableTabs.indexOf(props.activeTab) >= 0) {
       states.activeTab = props.activeTab;
     }
 
@@ -47,7 +52,7 @@ class SearchTabGroup extends React.Component {
   }
 
   renderTabControls() {
-    let tabControls = this.state.availableTabs.map(tabName => {
+    let tabControls = this.availableTabs.map(tabName => {
       let classnames = classNames(`btn btn-tab`, {
         active: this.state.activeTab === tabName
       });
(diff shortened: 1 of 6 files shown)


components/search-tab-group/search-tab-group.jsx
```diff
@@ -9,26 +9,31 @@ import ProfileLoader from "../profile-loader/profile-loader.jsx";
...
+// this component to re-render only when there's change in its props or state.
+// Note that React.PureComponent runs *shallow comparison* to determine
+// if the component should be re-rendered or not. So make sure
+// SearchTabGroup's props and state doesn't contain anything mutatable otherwise
+// we will run into issues.
+class SearchTabGroup extends React.PureComponent {
   constructor(props) {
     super(props);
 
+    this.availableTabs = TAB_NAMES;
```
mmmavis: This shouldn't have been a `state` in the first place 😬 Fixing this now.

pages/search/search.jsx
```diff
@@ -21,14 +21,7 @@ class Search extends React.Component {
   }
 
   componentDidMount() {
-    // The focus() function of <input /> isn't exposed by <DebounceInput />
```
mmmavis: Avoid accessing/manipulating DOM directly in React unless you really don't have a way to access element within the React component

pages/search/search.scss
```diff
@@ -1,4 +1,8 @@
 .search-page {
+  .btn-dismiss {
```
mmmavis: There are so many custom/new button styles on Pulse... I don't know what the best way is to keep thing tidy anymore 😂

pages/search/search.scss
```diff
@@ -55,17 +59,13 @@
...
-    @media (min-width: $bp-md) {
-      top: 1.4rem;
-      right: 0.5rem;
-    }
+    top: 0;
+    bottom: 0;
+    right: 12px;
+    height: 20px;
+    width: 20px;
+    margin: auto;
```
mmmavis: vertically centering it

mmmavis:
pinging @natalieworth 😄 this is ready for design review

natalieworth:
This looks great @mmmavis

components/search-tab-group/search-tab-group.jsx
youriwims: Why shouldn't it have been a state?

pages/search/search.jsx
youriwims: Agreed, I didn't know that issue in your comment was closed...I should've checked.

pages/search/search.scss
youriwims: nice

components/search-tab-group/search-tab-group.jsx
mmmavis: We use `state` for variable whose values changes over the course of the component's lifetime. You can see in the file that `availableTabs" never gets changed so hence it shouldn't have been stored in `state`.

pages/search/search.jsx
```diff
@@ -101,24 +94,40 @@ class Search extends React.Component {
...
+
+  renderKeywordDimissButton() {
+    return (
+      <button
+        className={`btn btn-dismiss small p-0 ${
+          this.state.searchBoxFocused ? `dismiss-focus-glyph` : `dismiss-glyph`
+        }`}
+        onClick={() => this.handleKeywordDismissBtnClick()}
+      >
+        <span className="sr-only">Clear input</span>
```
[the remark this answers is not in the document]
mmmavis: I don't know what the pros and cons are between the two approaches. But using `sr-only` (a Bootstrap's way) is what we've been using in the codebase so I chose to stick with it. Do you wanna file a ticket for the team to chat about which approach we should stick with in the future?

components/search-tab-group/search-tab-group.jsx
mmmavis: We can rename this variable since "available" tabs can sound confusing and misleading.

pages/search/search.jsx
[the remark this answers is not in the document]
youriwims: Thanks for the idea. Actually just added it to the next engineering meeting agenda. I'll go ahead and approve :)

mmmavis:
Yay!

PR #1131, opened 2019-08-07
Add max width to images inside bio page

alanmoo:
Fixes issues like https://www.mozillapulse.org/profile/2099


components/bio/bio.scss
```diff
@@ -78,4 +78,7 @@
   hr {
     border-top: 1px solid $black;
   }
+  img {
```
alanmoo:
```suggestion
  
  img {
```

PR #1130, opened 2019-08-03
Entry Page Updates

youriwims:
Closes #1079

PR #1138, opened 2019-08-21
Entry button Fix

youriwims:
Related issue: #1130
Summary: Visit button spans full width when Get Involved button isn't there (see [here](https://pulse.mofostaging.net/entry/492)). Fix involves keeping Visit button half width even when Get Involved Button not visible.

p.s. sorry mavis 😭 

_Note_: Commit should say 'bug' fix...not 'big' fix.

### DIFF
--- a/components/project-card/project-card-detailed.jsx
+++ b/components/project-card/project-card-detailed.jsx
@@ -122,13 +122,17 @@ class DetailedProjectCard extends React.Component {
   }
 
   renderVisitButton() {
+    let classes = classNames("btn btn-primary mb-3 mb-md-0 mr-md-3 d-flex justify-content-center align-items-center", {
+    "single-btn": !this.props.getInvolvedUrl
+    });
+
     if (!this.props.contentUrl) return null;
 
     return (
       <a
         href={this.props.contentUrl}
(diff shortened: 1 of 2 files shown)


components/project-card/project-card-detailed.jsx
```diff
@@ -122,13 +122,17 @@ class DetailedProjectCard extends React.Component {
   }
 
   renderVisitButton() {
+    let classes = classNames("btn btn-primary mb-3 mb-md-0 mr-md-3 d-flex justify-content-center align-items-center", {
```
mmmavis: - move this to after `if (!this.props.contentUrl) return null;`
- fix indentation

components/project-card/project-card-detailed.scss
```diff
@@ -144,6 +144,12 @@
       @media (min-width: $bp-lg) {
         flex: 1;
       }
+
+      &.single-btn {
```
mmmavis: Merge this and above into

```scss
@media (min-width: $bp-lg) {
  flex: 1;

  &.single-btn {
    flex: 0.5;
  }
}
      
```

Issue #1159, opened 2019-11-19
[Epic] Update outdated npm packages


mmmavis:
The packages we use for Pulse are pretty outdated. Let's update them!

<img src="https://user-images.githubusercontent.com/2896608/69173903-95fd8680-0ab5-11ea-9987-0a889325ecf0.png" width=420>

- [x] autoprefixer (dependabot PR #1180)
- [x] babel-loader (PR #1183)
- [x] browserslist (PR #1173)
- [x] chokidar (PR #1165)
- [x] chokidar-cli (PR #1165)
- [x] dotenv (issue #1193)
- [x] eslint (PR #1164)
- [x] eslint-config-prettier (PR #1164)
- [x] eslint-plugin-prettier (PR #1164)
- [x] eslint-plugin-react (PR #1164)
- [x] event-stream (issue #1192)
- [x] express (issue #1194)
- [x] helmet (PR #1165)
- [x] mofo-bootstrap (issue #1195)
- [x] mofo-ui (dependabot PR #1185)
- [x] node-sass (issue #1195)
- [x] optipng-bin (PR #1173)
- [x] ~~postcss-cli-simple~~ replaced by postcss-cli (PR #1173)
- [x] qs (issue #1197)
- [x] react (PR #1184)
- [x] react-debounce-input (PR #1175)
- [x] react-dom (PR #1184)
- [x] react-formbuilder (dependabot PR #1168 )
- [x] react-ga (issue #1195)
- [x] react-markdown (PR #1184)
- [x] react-modal (dependabot PR #1169)
- [x] react-router (PR #1182)
- [x] react-router-dom (PR #1182)
- [ ] react-select (issue #1196)
- [x] react-tag-autocomplete (issue #1195)
- [x] shx (dependabot PR #1185)
- [x] stylelint (PR #1164)
- [x] stylelint-config-standard (PR #1164)
- [x] svgo (PR #1164)
- [x] webpack (PR #1165)
- [x] webpack-cli  (PR #1165)

alanmoo:
@mmmavis to convert this to an Epic

alanmoo:
@mmmavis to break into smaller tickets with points (instead of this epic having points) and add them to the milestone/epic

mmmavis:
I'm gonna close this epic since most of the work is done & we have the remaining task filed as its own ticket (update `react-select` #1196)

PR #1184, opened 2019-12-06
Update 'react', 'react-dom', and 'react-markdown' to latest version

mmmavis:
Related issue: #1159


package.json
```diff
@@ -97,13 +97,13 @@
...
-    "react": "^15.6.2",
+    "react": "^16.12.0",
     "react-debounce-input": "^3.2.2",
-    "react-dom": "^15.6.2",
+    "react-dom": "^16.12.0",
     "react-formbuilder": "^0.10.0",
     "react-ga": "^2.6.0",
     "react-helmet": "^5.2.1",
-    "react-markdown": "^3.6.0",
+    "react-markdown": "^4.2.2",
```
mmmavis: https://github.com/rexxars/react-markdown/blob/master/CHANGELOG.md#breaking

components/form-fields/help-types.jsx
```diff
@@ -7,7 +7,7 @@ class HelpTypesField extends React.Component {
     super(props);
     this.state = { options: [] };
   }
-  componentWillMount() {
+  componentDidMount() {
```
mmmavis: Swapped 'componentWillMount' with 'componentDidMount' as recommended by React

components/form-fields/issues.jsx
```diff
@@ -10,7 +10,7 @@ class IssuesField extends React.Component {
       selectedOptions: this.props.value
     };
   }
-  componentWillMount() {
```
mmmavis: Swapped 'componentWillMount' with 'componentDidMount' as recommended by React

pages/not-found.jsx
```diff
@@ -4,7 +4,7 @@ import { Link } from "react-router-dom";
 import HintMessage from "../components/hint-message/hint-message.jsx";
 
 class NotFound extends React.Component {
-  componentWillMount() {
+  componentDidMount() {
```
mmmavis: Swapped 'componentWillMount' with 'componentDidMount' as recommended by React.

components/profile-about-tab/profile-about-tab.jsx
```diff
@@ -10,16 +10,16 @@ class ProfileAboutTab extends React.Component {
   renderLongBio() {
     if (!this.props.longBio) return null;
 
-    return this.props.longBio.split(`\n`).map(line => {
-      return (
```
mmmavis: https://github.com/rexxars/react-markdown/blob/master/CHANGELOG.md#breaking (2nd point)

> On React >= 16.2, if no className prop is provided, a fragment will be used instead of a div. To always render a div, pass 'div' as the root renderer.

package.json
```diff
@@ -97,13 +97,13 @@
     "postcss-cli": "^6.1.3",
     "prop-types": "^15.7.2",
     "qs": "^6.7.0",
-    "react": "^15.6.2",
+    "react": "^16.12.0",
```
mmmavis: https://github.com/facebook/react/blob/master/CHANGELOG.md#breaking-changes

components/profile-about-tab/profile-about-tab.jsx
Pomax: Do we _need_ a div here, though? I assume `ReactMarkDown` already generates everything wrapped in a master element?

mmmavis: Just did some test. It seems like every non-heading element is wrapped in `<p>` by default. I'll remove `<div>` from here.

components/form-fields/help-types.jsx
mmmavis: We can't move `Service.helpTypes.get()...` to `constrcutor` as the call is async and we are not allowed to update `state` in `constructor`.

components/form-fields/issues.jsx
mmmavis: Same reason.

We can't move `Service.issues.get()...` to `constructor` as the call is async and `this.state = { ... }` needs to be set synchronously within `constructor`.

pages/not-found.jsx
```diff
@@ -3,34 +3,30 @@ import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
 import HintMessage from "../components/hint-message/hint-message.jsx";
 
-class NotFound extends React.Component {
```
mmmavis: Changed into a stateless component. See code diff without whitespace changes: https://github.com/mozilla/network-pulse/pull/1184/files?w=1#diff-ec29f66da11909216aa5a68ff0259763

mmmavis:
@Pomax this is ready for review again!

components/form-fields/help-types.jsx
mmmavis: What I said above is not true. Because

> Service.get() immediately returns a value (namely, a Promise) and code execution continues, so the constructor doesn't wait for Service.helpTypes.get().then(...) to finish.

I've updated the PR. 🙈

components/form-fields/issues.jsx
mmmavis: What I said above is not true. Because

> Service.get() immediately returns a value (namely, a Promise) and code execution continues, so the constructor doesn't wait for Service.issues.get().then(...) to finish.

I've updated the PR. 🙈

PR #1246, opened 2020-02-21
Pass github token in headers and not as a query param

patjouk:
Using token in query params are not deprecated.

Related PRs/issues mozillafoundation/mofo-devops-private#297


js/review-app-slack-webhook.js
```diff
@@ -18,7 +18,12 @@ const slack_webhook = process.env.SLACK_WEBHOOK;
 if (pr_number) {
   request(
     `https://api.github.com/repos/${org}/${repo}/pulls/${pr_number}&access_token=${github_token}`,
```
cadecairos:
```suggestion
    `https://api.github.com/repos/${org}/${repo}/pulls/${pr_number}`,
```

lets remove that query param

PR #1247, opened 2020-02-24
Removed 'request' as dependency

### LINKED ISSUE #1245: remove `request` dependency
We rely on `request`, at least according to our dependency list, but https://www.npmjs.com/package/request is deprecated and will never be updated ever again. We should find out where we're using it, and replace it with modern code =)

/cc @mmmavis

mmmavis:
Closes #1245 

Replaced `request` by node's build-in `https` + Promise.

@patjouk can you verify that my PR doesn't break the script?

### DIFF
--- a/js/review-app-slack-webhook.js
+++ b/js/review-app-slack-webhook.js
@@ -17,7 +17,9 @@ const slack_webhook = process.env.SLACK_WEBHOOK;
 // For review app manually created, we have to use the branch name instead.
 
 const request = (url, options) => {
-  console.log(`\n\n\n====================\n[new request] ${url}\n====================\n\n\n`);
+  console.log(
+    `\n\n\n====================\n[new request] ${url}\n====================\n\n\n`
+  );
   return new Promise((resolve, reject) => {
     https
       .get(url, options, res => {
@@ -79,13 +81,11 @@ let slack_payload = {
 };
 
 if (pr_number) {
-  request(
-    `https://api.github.com/repos/${org}/${repo}/pulls/${pr_number}&access_token=${github_token}`,
-    {
-      method: "GET",
-      headers: { "User-Agent": "request" }
-    }
-  )
+  request(`https://api.github.com/repos/${org}/${repo}/pulls/${pr_number}`, {
+    method: "GET",
+    headers: { "User-Agent": "request" },
+    Authorization: `token ${github_token}`
+  })
     .then(gBody => {
       const pr_title = JSON.parse(gBody)["title"];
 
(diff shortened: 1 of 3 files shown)


mmmavis:
@patjouk I'd like to get some help here. Tried everything I could think of but it's still giving me `400` back when posting to Slack. I've updated my PR to print more logs for debugging purpose. Could you lend a hand here?

js/review-app-slack-webhook.js
```diff
@@ -23,8 +23,10 @@ const request = (url, options) => {
         let body = "";
         res.on("data", chunk => (body += chunk));
         res.on("end", () => {
-          if (res.statusCode !== 200) {
-            reject("Status code is not equal to 200.");
+          console.log(res);
```
Pomax: rogue console log?

patjouk:
`request` is calling `https.get` that forces `method` [to `GET`](https://nodejs.org/api/http.html#http_http_get_url_options_callback). Since you're using `request` in `postToSlack` to do a `POST` request, it won't work.
You could use `https.request` instead, but you won't be able to pass `json` [in options](https://nodejs.org/api/http.html#http_http_request_options_callback).

I remember getting frustrated by HTTPS and decided at some point to switch to requests. Since we're moving away from it, maybe you want to give [bent](https://github.com/mikeal/bent) a try?

Pomax:
If we're going to add a library, I'd probably recommend node-fetch, which gives adds the standards Fetch API to node and will allow a nice and clean `fetch(url, { method: 'POST', ...}).then(,,,)`

cadecairos:
If we want to use `https` to post data it's a bit involved:

```
const https = require('https')

const data = JSON.stringify({
  data: 'foo'
})

const options = {
  hostname: 'api.example.com',
  port: 443,
  path: '/data',
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    'Content-Length': data.length
  }
}

const req = https.request(options, (res) => {
  console.log(`statusCode: ${res.statusCode}`)

  res.on('data', (d) => {
    process.stdout.write(d)
  })
})

req.on('error', (error) => {
  console.error(error)
})

req.write(data)
req.end()
```

mmmavis:
Omg!!!! `https.get`!!!!! I remember I changed it into `https.request` during my first commit omg how could I not catch this!!!!!!!!

mmmavis:
Thanks @patjouk @Pomax @cadecairos ! I'll try to stick with using the built-in `https` module. If things get too complicated I'll try `fetch`

mmmavis:
Yay! Got this working!

mmmavis:
@Pomax @patjouk please give this PR another look!

mmmavis:
Yay merging!

PR #1264, opened 2020-03-14
Use 'heroku-postinstall' instead of 'postinstall'

### LINKED ISSUE #1191: Replace `postinstall` with `heroku-postbuild`
Follow-up for PR #1165. See [thread](https://github.com/mozilla/network-pulse/pull/1165/files#r351076602) for context.

@patjouk I set up `postinstall: npm run build` years ago when I started this project. I had it mainly for Heroku apps to run the build task so the site can be loaded properly (I guess I couldn't find a better way at the time?). In the PR above I split webpack configs into "production" version and "dev" version & change `postinstall` to `postinstall: npm run build:prod` so my PR doesn't break Heroku apps. However, we don't have devs to accidentally run the app in `production` mode just because they run `npm i` locally. Question: can we get rid of `postinstall` completely and use `heroku-postbuild` (or other heroku script) instead?

mmmavis:
Fixes #1191

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ We love contributors, but the team maintaining this project is small and not str
 $> git clone https://github.com/mozilla/network-pulse.git
 $> cd network-pulse
 $> npm install
+$> npm start
 ```
 
 If you would like to override default environment variables... create a `.env` file on the root directory and set your env vars there. See [environment variables section](https://github.com/mozilla/network-pulse#environment-variables) for details.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,6 +25,7 @@ We love contributors, but the team maintaining this project is small and not str
 $> git clone https://github.com/mozilla/network-pulse.git
 $> cd network-pulse
 $> npm install
+$> npm start
```
mmmavis: `npm start` runs the app on dev mode.

PR #1275, opened 2020-03-27
Limit image size to 500kb

patjouk:
Related PR: mozilla/network-pulse-api#628

I'm changing the image size limit to 500kb. Pulse homepage is 12MB of images, and profiles [like this one](https://www.mozillapulse.org/profile/82) are monsters of 58MB of images.

As a follow-up, I want to convert all existing images to respect that new limit. The ticket is here: https://github.com/mozilla/network-pulse-api/issues/629


mmmavis:
Re: adding hint for users

Yes I was thinking about the same. Let me see if I can add it in quickly. If it takes too much effort, I'll file a new ticket for it.

mmmavis:
@Pomax can you give this PR a quick look again?

PR #1504, opened 2021-01-14
(WIP) Clean up existing Google Analytics code & enable Google Tag Manager

mmmavis:
# ⚠️ WIP

Related issue: https://github.com/mozilla/foundation.mozilla.org/issues/5976

PR #1505, opened 2021-01-14
GA improvements with respect to DNT setting & install GTM

mmmavis:
Related issue: https://github.com/mozilla/foundation.mozilla.org/issues/5976

My [first attempt](https://github.com/mozilla/network-pulse/pull/1504) (mimicking the code we have on Foundation site) didn't end well. The two codebase setups are very different and that the majority of the Pulse code is run both server and client side. Plus on Foundation site, `google-analytics.js` and `react-ga-proxy.js` import each other and Pulse doesn't like that circular dependency. I ended up porting the same code to Pulse's [`analytics.js`](https://github.com/mozilla/network-pulse/pull/1505/files#diff-c414ed6c85094401db43ae1a56f0e0a118e6ef95ecfe25e3259d5c395adc6e20) as it's the most straightforward way to go.


js/analytics.js
```diff
@@ -1,32 +1,81 @@
...
+
+  // A no-operation object with the same API surface as ReactGA
+  Analytics.ReactGA = {
+    initialize: () => {},
+    event: () => {},
+  };
+  Analytics.initialize = () => {};
+} else {
+  // script is being run client-side
+
```
Pomax, line 16: I'd recommend restructuring the script a little so that we don't even build the real analytics object unless DNT is off, with the DNT checking function made a top level function:

```js
function checkDoNotTrack() {
  // Treat server-side prerender as "do not track". If JS is enabled, the
  // React app will load client-side and add in GA later if permitted.
  if (typeof window === `undefined`) return true;

  // ...rest of code here...
}

// Are we allowed to track?
if (checkDoNotTrack() === false) {
  // build our analytics export
  const Analytics = {
    ReactGA,
    initialize: function() {
      // ...
    }
  };

  export default Analytics;
}

// If we're not allowed to track, return a "no-op" API object.
else {
  function noop() {};
  
  const NoAnalytics = {
    ReactGA: {
      initialize: noop,
      event: noop,
    },
    initialize: noop
  };

  export default NoAnalytics
}
```

Pomax (changes requested):

js/analytics.js
mmmavis: Nice suggestion!

mmmavis:
PR updated so analytics.js is cleaner!

Pomax (approved):
so many renames~ =D

PR #1540, opened 2021-03-04
Add noindex, nofollow, noarchive for staging server

### LINKED ISSUE #1529: Staging Sites don't have no-index
Pulse staging doesn't have a no-index meta tag or robots.txt so it is showing up in google. This can be harmful as google may detect Pulse or Pulse staging as duplicate content and it may negatively impact our SEO.

See: https://moz.com/learn/seo/duplicate-content for more info

* [ ] update build process to automatically include a no-index, whether it be through a meta tag, robots.txt or other.
* [ ] verify it doesn't apply to the live site
* [ ] check our other properties to make sure the same problem doesn't exist elsewhere and make tickets to fix if it does

Pomax:
Closes #1529

This adds noindex/nofollow/noarchive to our express responses as both an HTTP header and a google-specific meta tag (because of course Google is special, why wouldn't it be).

This relies on an env var `STAGING_SERVER`, which I've already set to `True` on the heroku staging server.

### DIFF
--- a/server.js
+++ b/server.js
@@ -46,7 +46,7 @@ app.use(
 );
 
 // No-index and No-follow in "not production"
-if (STAGING_SERVER) {
+if (env.STAGING_SERVER) {
   app.use((req, res, next) => {
     res.setHeader('X-Robots-Tag', 'noindex, noarchive, nofollow');
     next();
--- a/server.js
+++ b/server.js
@@ -48,7 +48,7 @@ app.use(
 // No-index and No-follow in "not production"
 if (env.STAGING_SERVER) {
   app.use((req, res, next) => {
-    res.setHeader('X-Robots-Tag', 'noindex, noarchive, nofollow');
+    res.setHeader("X-Robots-Tag", "noindex, noarchive, nofollow");
     next();
   });
 }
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -45,6 +45,14 @@ app.use(
   })
 );
 
+// No-index and No-follow in "not production"
+if (env.STAGING_SERVER) {
```
mmmavis, line 49: let's turn this into a `const`, e.g.,

```js
const ON_STAGING_SERVER = env.STAGING_SERVER === `True`;
```

and use `ON_STAGING_SERVER` instead.

server.js
```diff
@@ -129,6 +137,11 @@ function renderPage(appHtml, reactHelmet, canonicalUrl) {
         <title>"${meta.title}"</title>
         <meta name="description" content="${meta.description}">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        ${
+          env.STAGING_SERVER && (
```
mmmavis, line 141: Unfortunately this `${ boolean && string}` syntax doesn't work here. It will also output true/false as string as part of the DOM 😞

mmmavis (changes requested):
image

Wow I totally forgot we have this useful documentation checker on Pulse!

server.js
Pomax: It shouldn't, because `true && something` is `something` in JS O_o

I'll have a look at what's going on because I use that pattern all over the place in side projects and always see it do the right thing =S

mmmavis: I thought so too, but it outputs (see screenshot) on the screen if I set the env var to `True`

image

I haven't checked its source code but I wonder if it's `react-dom/server` doing some weird conversion on their end.

mmmavis: Ah found the issue! 😂 We just need to wrap `<meta name="googlebot" content="noindex, nofollow, noarchive" />` in quotes since we are dealing things in strings and not React nodes here.

mmmavis (changes requested):
Just need to wrap a line in quotes! 😉

server.js
Pomax: haha, omg well spotted =D

mmmavis (approved):
Yay 🎉

PR #1637, opened 2022-01-25
Add recaptcha

Pomax:
This PR covers the client-side recaptcha work outlined in https://github.com/mozilla/network-pulse-api/issues/762

This gates the login route behind recaptcha validation, if USE_RECAPTCHA is enabled.

Requires https://github.com/mozilla/network-pulse-api/pull/766 on the API-side

# Testing

- start with the test instructions on https://github.com/mozilla/network-pulse-api/pull/766
- clone the repo
- make sure you have Node v16 installed, or if you (wisely) use `nvm`, set it to use v16 for the duration of testing. v17 upgraded its SSL library in a way that broke a million codebases and making things Node 17 compatible is a _whole_ other can of worms.
- run `npm install`
- create an `.env` file and set it to:
```
PULSE_API=http://localhost:8000/api/pulse
PULSE_API_HOST=http://localhost:8000
PULSE_LOGIN_URL=http://localhost:8000/login
USE_RECAPTCHA=True
RECAPTCHA_KEY=therecaptchasitekey
```
where the reCAPTCHA site key is the `localhost (v3)` site key found on https://www.google.com/recaptcha/admin
- run `npm start`

You should now be able to load up http://localhost:3000 and click the signin/signup link in the nav bar. Doing so should show you print statements pertaining to recaptcha in your API server terminal fly by. If all went well, you should now be asked to pick how you want to authenticate. Note: that is all you need to see. Actually logging in won't work (due to the gsuite callback URL not being a localhost URL for obvious reasons =D)

Please play around with what happens with `USE_RECAPTCHA` turned off only on the client, only on the server, and off for both: this should fail, pass, and pass respectively.


danielfmiranda (approved):
Hi Pomax, just tested this alongside the other PR and can confirm that I get the expected results when using recaptha and also when turning it off for one app or the other, as well as both at the same time.

Approving!

tbrlpld:
I was surprised to find that `USE_RECAPTCHA=False` does not deactivate recaptcha. It still appears to work. Removing the key from the `.env` file does have the intended effect.

tbrlpld:
I can also confirm the expected behaviour

tbrlpld (approved):

PR #1657, opened 2022-08-09
 updated to use data layer + only fire when checkbox is checked

danielfmiranda:
Related issue:#1656

This PR adds a new data layer event to the pulse newsletter sign up form at the footer of the app, as well as moves the code within an if statement, to make sure that the event doesnt fire unless the user selects the privacy consent checkbox.

**Note:**

Per instructions in #1656, I have removed the language and country parameters, since the sign up form in pulse does **not** use these.
```
dataLayer code:
// Mozilla Foundation + Torchbox - 2022
// dataLayer push for newsletter signup form submission

window.dataLayer = window.dataLayer || [];
window.dataLayer.push({
  event: 'form_submission',
  form_type: 'newsletter_signup', // type of form. should be static in this case
  form_location: 'header', // update with location of form in page -- either 'header' or 'footer', depending on which form the user used to sign up for the newsletter
  country: 'argentina', // update according to the country the user selected from the dropdown, if the form has a country dropdown
  language: 'english' // update according to the language that the user selected from the dropdown, if the form has a language dropdown
  }
);

```

**Screenshot of event firing when its supposed to:**
image

### DIFF
--- a/components/newsletter-sign-up/newsletter-sign-up.jsx
+++ b/components/newsletter-sign-up/newsletter-sign-up.jsx
@@ -82,13 +82,21 @@ class NewsletterSignUp extends React.Component {
 
     if (email && this.validatesAsEmail(email) && consent) {
       this.submitDataToApi();
+      Analytics.ReactGA.event({
+        category: `signup`,
+        action: `form submit tap`,
+        label: `Signup submitted`,
+      });
+      window.dataLayer = window.dataLayer || [];
+      window.dataLayer.push({
+        event: 'form_submission',
+        form_type: 'newsletter_signup', // type of form. should be static in this case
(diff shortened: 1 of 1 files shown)


components/newsletter-sign-up/newsletter-sign-up.jsx
```diff
@@ -82,13 +82,18 @@ class NewsletterSignUp extends React.Component {
 
     if (email && this.validatesAsEmail(email) && consent) {
       this.submitDataToApi();
+      Analytics.ReactGA.event({
```
danielfmiranda, line 85: Moved this within the if statement instead of outside, that way this will only run if the email is filled out as well as the consent box!

fessehaye (approved):

